PR #9, opened 2018-02-19
ensureObject

mymindstorm:
Diff is weird because of gts fix doing the indenting this time. Last time storage.ts was changed that wasn't synced to my branch was 67a42c9

### DIFF
--- a/src/models/storage.ts
+++ b/src/models/storage.ts
@@ -23,6 +23,9 @@ class EntryStorage {
     const tempEntryArray: OTPStorage[] = [];
 
     for (const hash of Object.keys(_data)) {
+      if (!this.isValidEntry(_data, hash)) {
+        continue;
+      }
       tempEntryArray.push(_data[hash]);
     }
 
@@ -39,12 +42,33 @@ class EntryStorage {
     return newData;
   }
 
-  private static ensureObject(
+  private static isOTPStorage(entry: object) {
+    const properties = [
+      'account', 'hash', 'index', 'issuer', 'type', 'counter', 'secret',
+      'encrypted'
+    ];
+    for (let i = 0; i < properties.length; i++) {
+      if (!entry.hasOwnProperty(properties[i])) {
+        return false;
+      }
+    }
+    return true;
+  }
+
+  private static isValidEntry(
       _data: {[hash: string]: OTPStorage}, hash: string) {
+    if (!_data.hasOwnProperty(hash)) {
+      console.log('Key "' + hash + '" does not exist');
+      return false;
+    }
     if (typeof _data[hash] !== 'object') {
       console.log('Key "' + hash + '" is not an object');
       return false;
     } else {
+      if (!this.isOTPStorage(_data[hash])) {
+        console.log('Key "' + hash + '" is not OTPStorage');
+        return false;
+      }
       return true;
     }
   }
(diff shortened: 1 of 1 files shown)


src/models/storage.ts
```diff
@@ -39,14 +39,26 @@ class EntryStorage {
     return newData;
   }
 
+  private static ensureObject(
+      _data: {[hash: string]: OTPStorage}, hash: string) {
+    if (typeof _data[hash] !== 'object') {
```
Sneezry: `_data[hash]` may not exist (we cannot ensure `hash` is gotten from `hash of Object.keys(_data)`, though we do it currently), `hasOwnProperty` should be a good check here.

src/models/storage.ts
```diff
@@ -62,15 +74,17 @@ class EntryStorage {
...
-                // decrypt the data to export
-                _data[hash].secret = _data[hash].encrypted ?
-                    encryption.getDecryptedSecret(_data[hash].secret) :
-                    _data[hash].secret;
-                _data[hash].encrypted = false;
-                // we need correct hash
-                if (hash !== _data[hash].hash) {
-                  _data[_data[hash].hash] = _data[hash];
-                  delete _data[hash];
+                if (this.ensureObject(_data, hash)) {
```
Sneezry: we can just skip the loop if ensureObject is false:
```
if (!this.ensureObject(_data, hash)) {
    continue;
}
```

Also, `ensureObject` is not a good name here, the name makes me think that the method will fix bad data issue, as what `ensureUniqueIndex` does, but it just returns check result. Suggestion: call it `isValidEntry`.

src/models/storage.ts
Sneezry: Also, we may check more here. `_data[hash]` is not only an `object`, but also should be `OTPStorage`. So we can check if it has the required fields for `OTPStorage`, such as `type`, `secret`, etc.

src/models/storage.ts
```diff
@@ -193,77 +207,79 @@ class EntryStorage {
             chrome.storage.sync.get((_data: {[hash: string]: OTPStorage}) => {
               const data: OTPEntry[] = [];
               for (let hash of Object.keys(_data)) {
-                const entryData = _data[hash];
-                let needMigrate = false;
+                if (this.ensureObject(_data, hash)) {
```
Sneezry: Same here, we can just skip the loop if check not passes

src/models/storage.ts
mymindstorm: I wouldn't call it `isValidEntry` unless it checks if it's an OTPEntry too, I should actually check that too

src/models/storage.ts
mymindstorm: On the loops, I did it like this so that it just ignores the invalid data and keeps looping so that it will continue to process valid data. If I just have it break the loop won't it stop loading good data as soon as it runs into data it doesn't understand?

src/models/storage.ts
Sneezry: It should be `OTPStorage`, I messed the type. Maybe `isValidOTPStorage`?

Sneezry: Or is this method just for checking object, but not for `OTPStorage` checking?

mymindstorm: It is `OTPStorage`, oops. `isValidEntry` should be fine as long as I make it check everything you suggested, `isValidOTPStorage` is a bit long if it should be called that then `isOTPStorage` should be enough.

src/models/storage.ts
Sneezry: We do not `break` the loop here, but skip by using `continue` ;-P

mymindstorm: How did I not know about that >_<

src/models/storage.ts
```diff
@@ -39,12 +42,42 @@ class EntryStorage {
     return newData;
   }
 
+  private static isOTPStorage(entry: object) {
+    const properties = [
+      'account', 'hash', 'index', 'issuer', 'type', 'counter', 'secret',
```
mymindstorm, line 47: I feel like I shouldn't have to manually type this out, but apparently when ts is compiled it loses all type definitions

Sneezry: It's okay. TypeScript is syntax sugar for JavaScript, it cannot handle everything. Hardcode here is fine.

Sneezry (approved):

PR #11, opened 2018-02-19
midify i18n message

Sneezry:


mymindstorm:
If you think that secret isn't a good word to use we could look into changing it to 'code' or something more user-friendly

Sneezry:
secret should be fine, what's your opinion?

mymindstorm:
I don't really think that it's used enough to matter. We use 'secret' in context of an action in this case anyway so it should be very obvious what it means.

PR #37, opened 2018-02-23
Overflow fix

### LINKED ISSUE #36: l18n strings overflow

mymindstorm:
Why is it so hard to center things?

closes #36

### DIFF
--- a/css/popup.css
+++ b/css/popup.css
@@ -555,7 +555,8 @@ body {
     font-size: 16px;
     color: gray;
     cursor: pointer;
-    text-overflow: ellipsis;
+    display: grid;
+    grid-template-columns: 30px auto;
 }
 
 #menu .menuList p:hover {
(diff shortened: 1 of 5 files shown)


Sneezry:
The menu looks nice now. The Dropbox code input box looks not as good as previous, how about to change it back to previous version?

css/popup.css
```diff
@@ -515,10 +515,17 @@ body {
     height: 480px;
     position: absolute;
     left: -1000px;
-    background: #EEE;
     top: 0;
 }
 
+#menuBody {
+    overflow-y: scroll;
```
Sneezry: `auto` should be better here

mymindstorm:
I think I was able to make it cooperate.

Sneezry:
Awesome!

Sneezry (approved):

PR #38, opened 2018-02-24
Add localStorage password warning

mymindstorm:
This does not work. I assume it's because this runs before `_ui.instance` is created, so I don't really know where a good place to put this check is instead of here.

### DIFF
--- a/src/popup.ts
+++ b/src/popup.ts
@@ -31,7 +31,8 @@ async function init() {
   }
 
   // localStorage passphrase warning
-  if (localStorage.encodedPhrase !== null) {
+  console.log(typeof localStorage.encodedPhrase);
+  if (localStorage.encodedPhrase) {
     authenticator.alert(authenticator.i18n.local_passphrase_warning);
   }
 
--- a/src/popup.ts
+++ b/src/popup.ts
@@ -30,6 +30,11 @@ async function init() {
     authenticator.showInfo('passphrase');
   }
 
+  // localStorage passphrase warning
+  if (localStorage.encodedPhrase !== null) {
+    authenticator.alert(authenticator.i18n.local_passphrase_warning);
+  }
+
   // Remind backup
   const backupReminder = setInterval(() => {
     if (authenticator.entries.length === 0) {
--- a/src/popup.ts
+++ b/src/popup.ts
@@ -31,7 +31,6 @@ async function init() {
   }
 
   // localStorage passphrase warning
-  console.log(typeof localStorage.encodedPhrase);
   if (localStorage.encodedPhrase) {
     authenticator.alert(authenticator.i18n.local_passphrase_warning);
   }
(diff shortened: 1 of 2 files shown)


Sneezry:
`_ui` shouldn't be called external. Should we enforce users to change the password? I think that's not friendly.

mymindstorm:
> `_ui` shouldn't be called external.

Not exactly sure what you mean by that.

> Should we enforce users to change the password? I think that's not friendly.

This isn't exactly enforcing, but I'd actually rather do something unfriendly since storing passwords in cookies and localStorage is a big no-no. On load of the page it should check if pass is in localStorage and either:

- Annoy the user with a message asking to update password every time the check runs
- Annoy the user with a yes / no prompt that clears localStorage on yes
- Remove the password from localStorage and notify the user once

I'm currently trying to implement option 1 here, since it'll cause the least amount of confusion. A yes / no prompt will probably just scare and / or confuse users. Just removing it will cause problems since some people will be locked out of their accounts.

Sneezry:
>> `_ui` shouldn't be called external.

> Not exactly sure what you mean by that.

You can regard `_ui` as a private member of UI config, so you shouldn't pass it to external functions, but use it within `UIConfig.methods` functions. `_ui.instance` points to the Vue instance itself, such as `authenticator`, so `_ui.instance` doesn't exist until we create the instance by `new Vue(ui_config)`.

I suggest to check the cached passphrase in localStorage before backup remind (before this line: https://github.com/Authenticator-Extension/Authenticator/blob/730bd48e44fbb91fd1a01372d93e3878fc5bfb1b/src/popup.ts#L33)

The logic between backup remind and clear local cached passphrase are almost the same, we check some thing when popup opens, and shows a message if necessary.

mymindstorm:
That makes sense to me, thanks! Do you have any better ideas about how to move users away from `localStorage` without being too unfriendly?

mymindstorm:
How does this look @Sneezry

mymindstorm:
Wait I did something wrong, its always true

Sneezry:
We can just check `localStorage.encodedPhrase`

```
if (localStorage.encodedPhrase) {
   // bla bla bla
}
```

if `localStorage.encodedPhrase` is not set, its value is `undefined`, but not `null`.

mymindstorm:
I'm just going to make it check for a string type then

src/popup.ts
```diff
@@ -30,6 +30,12 @@ async function init() {
     authenticator.showInfo('passphrase');
   }
 
+  // localStorage passphrase warning
+  console.log(typeof localStorage.encodedPhrase);
```
Sneezry: should we keep the debug info here?

Sneezry (approved):

src/popup.ts
mymindstorm: oops, forgot about that

PR #45, opened 2018-02-25
Message tweak

### LINKED ISSUE #44: Message windows need css tweaks

Message windows should have a border and like everything else.

mymindstorm:
closes #44 

Not sure about the shadow.

### DIFF
--- a/css/popup.css
+++ b/css/popup.css
@@ -776,7 +776,8 @@ body {
     position: absolute;
     width: 300px;
     padding: 10px;
-    border: gray;
+    border: gray 1px solid;
+    border-radius: 2px;
     background: white;
     left: 10px;
     top: 150px;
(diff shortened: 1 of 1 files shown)


mymindstorm:
Here's the window at 0.25 opacity if you want to try that:
image

css/popup.css
```diff
@@ -852,7 +853,7 @@ body {
     position: absolute;
     width: 100%;
     height: 100%;
-    background: rgba(0, 0, 0, 0.5);
+    background: rgba(0, 0, 0, 0);
```
Sneezry: we can safely remove this line as the default background is transparent

Sneezry (approved):

mymindstorm:
On the background what do you think transparency should be set to? 0.5 is
definitely too dark, but do you think that 0.25 would look better than none?

Sneezry:
transparent is okay. `background: rgba(0, 0, 0, 0);` equals to removing `background` attribute, so we can just remove it from the code.

PR #50, opened 2018-02-28
Stop caching passphase in cookies

mymindstorm:
Storing in cookies is bad and makes removing localStorage password storage pointless. I don't think there is a point anyway as long as its stored here:
https://github.com/Authenticator-Extension/Authenticator/blob/869ad1bd7a554b44e191e8763a48c94cb522b01d/src/background.ts#L17-L20

### DIFF
--- a/src/ui/entry.ts
+++ b/src/ui/entry.ts
@@ -89,17 +89,12 @@ function hasMatchedEntry(currentHost: string, entries: OTPEntry[]) {
 async function getCachedPassphrase() {
   return new Promise(
       (resolve: (value: string) => void, reject: (reason: Error) => void) => {
-        const cookie = document.cookie;
-        const cookieMatch =
-            cookie ? document.cookie.match(/passphrase=([^;]*)/) : null;
-        const cachedPassphrase =
-            cookieMatch && cookieMatch.length > 1 ? cookieMatch[1] : null;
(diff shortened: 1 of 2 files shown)


Sneezry:
cookie will be removed immediately after browser closed. Absolute security means hard to use, 1Password has similar behavior

mymindstorm:
Doesn't that mean it will last just as long as `cachedPassphrase`?

mymindstorm:
Oh I see, the event page closes after a few minutes.

mymindstorm:
@Sneezry what if we open a message port (to keep it from closing) in background.ts if a user password is set & useragent is chrome? Or just dropping `"persistent": false` entirely (runs at about 0.03 GB memory when loaded)?

Sneezry:
I think we have no need to do that. Cookie has the same life time as browser (not event page or popup page), and it makes sense. Event page and popup only run when they should.

Entering password every time will make the extension hard to use.

mymindstorm:
You won't have to enter password every time, but only if `background.ts` stays loaded. It would have the same lifetime as a cookie, but without the possibility of  getting read by malicious extensions. Still better than being written to disk, though.

Sneezry:
cookie under chrome-extension: should be safe

PR #58, opened 2018-03-04
Attempt to repair OTPStorage with missing keys

### LINKED ISSUE #57: Updating firefox version clears codes

Version disabled, investigating cause.

mymindstorm:
This attempts to re-initialize missing values from old version.

fixes #57

### DIFF
--- a/src/models/interface.ts
+++ b/src/models/interface.ts
@@ -36,6 +36,14 @@ interface OTPStorage {
   counter: number;
 }
 
+interface OTPTestObject {
+  account?: string;
+  hash?: string;
+  index?: number;
+  issuer?: string;
+  counter?: number;
+}
+
 /* tslint:disable-next-line:interface-name */
 interface I18nMessage {
   [key: string]: {message: string, description: string};
--- a/src/models/storage.ts
+++ b/src/models/storage.ts
@@ -42,14 +42,40 @@ class EntryStorage {
     return newData;
   }
 
-  private static isOTPStorage(entry: object) {
+  private static ensureOTPStorage(hash: string, entry: OTPTestObject) {
     const properties = [
       'account', 'hash', 'index', 'issuer', 'type', 'counter', 'secret',
       'encrypted'
     ];
     for (let i = 0; i < properties.length; i++) {
       if (!entry.hasOwnProperty(properties[i])) {
-        return false;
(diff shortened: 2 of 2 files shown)


Sneezry:
Oh, it seems we need a hotfix. The logic seems not correct, it only fix one property in the loop. As this is a hotfix, I will handle this issue to fix it fast.

mymindstorm:
Its part of line 50's loop so it should loop through every property one at a time.

mymindstorm:
@Sneezry 

https://github.com/Authenticator-Extension/Authenticator/blob/ad7e4310460bc1aa4c7f7d4ba791ca07186cb90e/src/models/storage.ts#L50

it should loop just fine

Sneezry:
It is not only check valid issue, previous migrate logic is incorrect either. I have pushed the hotfix https://github.com/Authenticator-Extension/Authenticator/pull/59

Sneezry:
Could you please post a wiki to tell user how to migrate data from old version to 5.0.4?

The user needs get raw data by `chrome.storage.sync.get(console.log)`, then copy the text and paste into import page.

mymindstorm:
I can, but make it open a link to it if it detects that a user has old data

Sneezry:
5.0.5 has been merged into release branch. I set 5.0.4 to push 1% user on Chrome for public preview, is there a similar option for Firefox market?

mymindstorm:
No, I disabled the release completely as soon as I learned about the issue. Firefox updates are really slow, so I doubt much of the userbase have gotten it yet.

mymindstorm:
So just so we are on the same page here, the old version was saving without some required keys. With your hotfix now users need to re-import? There isn't anything we can do to convert to new format without user intervention?

Sneezry:
Not really. `EntryStorage.get` will handle migration:

https://github.com/Authenticator-Extension/Authenticator/blob/e993945612c44a2091439ba989b8853ce697fe54/src/models/storage.ts#L263

mymindstorm:
Then we should get rid of 5.0.4 from distribution and ask any users that have problems with 5.0.4 to update to 5.0.5 instead of writing documentation?

Sneezry:
As 5.0.5 is set to push to 1% user as well, I'm not sure they can get the update. However, 5.0.5 has been released to Chrome Web Store, and it weekend today, we can wait to see if any user post issues.

PR #65, opened 2018-03-04
Fix character encoding issue

### LINKED ISSUE #63: Setting non-standard characters in account fields breaks extension
Why when I try to open the program can only see a white window, and the codes do not appear?

mymindstorm:
should fix #63


mymindstorm:
What do you think about this @Sneezry?

Sneezry:
Looks good. Does it work? If it works, we can merge it.

mymindstorm:
I tested and it worked, but you should test too when you have time.

Sneezry:
Sure

Sneezry:
It works on my side as well :-D

mymindstorm:
Great!

PR #66, opened 2018-03-05
5.0.6 dev to master

mymindstorm:
I'd like to push bugfix soon, anything else we need before pushing this?

### DIFF
--- a/manifest-chrome.json
+++ b/manifest-chrome.json
@@ -2,7 +2,7 @@
     "manifest_version": 2,
     "name": "__MSG_extName__",
     "short_name": "__MSG_extShortName__",
-    "version": "5.0.5",
+    "version": "5.0.6",
     "default_locale": "en",
     "description": "__MSG_extDesc__",
     "icons": {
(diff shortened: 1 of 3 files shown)


Sneezry:
Kindly remind: please use `Create a merge commit` rather than `Squash and merge` between `dev` and `master` branch, or our time will mess again.

snipaste20180305_103937

mymindstorm:
I'm pretty sure you can disable squash merging in settings

Sneezry:
I have no idea if I can disable squash merging on specific branch. When we need add new features, we create a new branch, and when the feature is ready, we squash merge it into `dev`, then delete the feature branch. But when we need merge `dev` into `master` or `master` into `release`, we need keep all commits. I can see merging settings in Merge button section, but it seems a global setting. Any idea?

mymindstorm:
Nope, just checked. I really thought they would have a way to disable that on protected branches.

Sneezry:
Not a big problem, we just know how to handle it now. Please merge master into release directly as well.

mymindstorm:
Done, release tagged too. Uploading to AMO

PR #69, opened 2018-03-09
Auto tag

### LINKED ISSUE #68: Auto release with Travis

mymindstorm:
closes #68


mymindstorm:
This also has the extra benefit of breaking the build if you don't update the version number in `manifest-chrome.json`

mymindstorm:
I think I might be able to make it automatically open pull requests after merging to master too

ci/tag.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+# This script is used by travis to auto tag our releases
+
+# Configure git
```
mymindstorm, line 4: If we want to do more `build` tasks w/ git then this section should be moved to another script file and run during `before_deploy`

Sneezry (approved):

PR #91, opened 2018-03-17
fix oauth flow

Sneezry:
We need know value of `chrome.identity.getRedirectURL()` in Firefox

closes #91

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -216,7 +216,7 @@
         "description": "Dropbox token."
     },
     "dropbox_authorization": {
-        "message": "Get Code",
+        "message": "Sign in Dropbox",
         "description": "Dropbox authorization."
     },
     "show_all_entries": {
@@ -234,5 +234,9 @@
     "local_passphrase_warning": {
         "message": "Your password is stored locally, please change it in the security menu immediately.",
         "description": "localStorage password warning."
+    },
+    "dropbox_logout": {
+        "message": "Logout Dropbox",
+        "description": "Dropbox logout."
     }
 }
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -216,7 +216,7 @@
         "description": "Dropbox token."
     },
     "dropbox_authorization": {
-        "message": "Sign in Dropbox",
+        "message": "Sign in to Dropbox",
         "description": "Dropbox authorization."
     },
     "show_all_entries": {
@@ -236,7 +236,7 @@
         "description": "localStorage password warning."
     },
     "dropbox_logout": {
-        "message": "Logout Dropbox",
+        "message": "Logout of Dropbox",
         "description": "Dropbox logout."
     }
 }
--- a/popup.html
+++ b/popup.html
@@ -143,7 +143,7 @@
                     <div id="dropbox_risk">{{ i18n.dropbox_risk }}</div>
                     <label for="token_input">{{ i18n.dropbox_token }}</label>
                     <input type="text" id="token_input" v-model="dropboxToken" disabled>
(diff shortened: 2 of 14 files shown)


mymindstorm (approved):
Works perfectly :D

The only big suggestions I have are:

- The UX for the "Get Code" button is a bit weird. We only use the smaller buttons for "Ok" buttons or to update data, not trigger an action. Maybe style it like`#download_backup`? 

- A string like "Login to Dropbox" would be more appropriate if you do make the button bigger. (i18n can use "Get Code" until other languages catch up)

- We need a button to clear the Dropbox token from storage labelled "Logout of Dropbox" or something similar. At minimum it should link to https://www.dropbox.com/account/connected_apps

- While I'm at it, does the "Dropbox Token" field serve a purpose now that we're using the proper auth flow?

_locales/en/messages.json
```diff
@@ -216,7 +216,7 @@
         "description": "Dropbox token."
     },
     "dropbox_authorization": {
-        "message": "Get Code",
+        "message": "Sign in Dropbox",
```
mymindstorm: "Sign in Dropbox" => "Sign in to Dropbox"

_locales/en/messages.json
```diff
@@ -234,5 +234,9 @@
     "local_passphrase_warning": {
         "message": "Your password is stored locally, please change it in the security menu immediately.",
         "description": "localStorage password warning."
+    },
+    "dropbox_logout": {
+        "message": "Logout Dropbox",
```
mymindstorm: "Logout Dropbox" => "Logout of Dropbox"

src/ui/info.ts
```diff
@@ -6,49 +6,12 @@ async function info(_ui: UI) {
...
-                }
-                return resolve();
-              });
-        });
+      getDropboxToken: () => {
+        chrome.runtime.sendMessage({action: 'dropbox'});
+      },
+      logoutDropbox: async () => {
+        localStorage.removeItem('dropboxToken');
+        _ui.instance.openLink('https://www.dropbox.com/account/connected_apps');
```
mymindstorm: Since line 13 removes the token it should be fine to remove opening this page. A success message would be useful too.

popup.html
```diff
@@ -143,7 +143,8 @@
                     <div id="dropbox_risk">{{ i18n.dropbox_risk }}</div>
                     <label for="token_input">{{ i18n.dropbox_token }}</label>
                     <input type="text" id="token_input" v-model="dropboxToken" disabled>
```
mymindstorm: Action buttons are more important than the token field. Buttons should be at the top of the page since they are what the users will be interacting w/, so it might be a good idea to move this down two lines.

mymindstorm (changes requested):

src/ui/info.ts
```diff
@@ -4,14 +4,16 @@
...
-    },
+    data: {info: '', dropboxToken: localStorage.dropboxToken || ''},
     methods: {
+      getDropboxToken: () => {
+        chrome.runtime.sendMessage({action: 'dropbox'});
+      },
+      logoutDropbox: async () => {
+        localStorage.removeItem('dropboxToken');
+        _ui.instance.dropboxToken = '';
+        _ui.instance.openLink('https://www.dropbox.com/account/connected_apps');
```
mymindstorm, line 15: I don't think we need to open page anymore since token is deleted from localstorage. It's not that important though, so you can merge if you think it should open this page.

mymindstorm (approved):
All good

src/ui/info.ts
Sneezry: I think user may not logout dropbox frequently, and if they click the button, they may just want to disconnect the extension from dropbox. So open the page make sense.

PR #92, opened 2018-03-17
Fix #77: QR code scanning on HiDPI screens

### LINKED ISSUE #77: QR code not scanning properly
Here's a sample barcode from the process:

index-gauth

nt1m:


Sneezry:
Emmm, it seems not make sense. `devicePixelRatio` is not 1 for retina display, and if we remove it from the code, scan QR should not work correctly. Have you tested the code?

nt1m:
@Sneezry Yes, I've tested the code, and it fixes #77.

The only moment where you need to be careful with `devicePixelRatio` is when manipulating the canvas API, but otherwise, you shouldn't scale dimensions by `devicePixelRatio`.

nt1m:
@Sneezry Do you need a screencast as proof that it works ?

Sneezry:
I will test it on macbook. Because the recent code works well on macbook.

That is really weird...

Sneezry:
It seems the change will break on MacBook, but we are clear that here's bug in HiDPI screen detection.

screen shot 2018-03-18 at 01 10 45

Could you please tell me the value of `window.devicePixelRatio` on your machine? And how about the resolution of your display? What is the value of `window.screen.width` on your machine?

It would be better if you can tell me your display model (such as DELL P4317Q).

nt1m:
I've tested my changes on my MacBook Pro 2016 15'. The difference however is that I've tested on Firefox, whereas you've tested in Chrome. There might be differences between the two browsers.

`window.devicePixelRatio` is 2.
screen shot 2018-03-18 at 12 37 52 am

nt1m:
With my pull request:

screen shot 2018-03-18 at 12 47 45 am

Without my pull request:

screen shot 2018-03-18 at 12 53 03 am

Sneezry:
I can repo this issue on Firefox. It seems Firefox capture screenshot with non-HiDPI image. Can you modify the code to make it work on both Chrome and Firefox?

Sneezry:
Hi @nt1m , I have made a new PR (https://github.com/Authenticator-Extension/Authenticator/pull/97) to make it work both on Chrome and Firefox. Please review the code if you'd like.

I will close this PR then. Thanks for your contribution!

PR #124, opened 2018-04-18
Fix #122

### LINKED ISSUE #122: Issuer name is not decoded properly if it contains a space
According to [Google Authenticator Key-URI format](https://github.com/google/google-authenticator/wiki/Key-Uri-Format#issuer), issuer name must be URL-encoded, but Authenticator-Extension doesn't properly decode it.

Steps to reproduce:
1. Consider the following totp URI, which has **ACME Co** as issuer name and corresponded QR code:
> otpauth://totp/ACME%20Co:john.doe@email.com?secret=HXDMVJECJJWSRB3HWIZR4IFUGFTMXBOZ&issuer=ACME%20Co
qr
2. Scan the QR code with Authenticator-Extension
3. Review issuer name of the added totp:
    - expected: **ACME Co**
    - actual: **ACME%20Co**

mymindstorm:


src/background.ts
```diff
@@ -94,7 +94,7 @@ async function getTotp(text: string, passphrase: string) {
         if (parameter[0].toLowerCase() === 'secret') {
           secret = parameter[1];
         } else if (parameter[0].toLowerCase() === 'issuer') {
-          issuer = parameter[1];
+          issuer = decodeURIComponent(parameter[1]);
```
Sneezry: use try catch here to prevent exception

mymindstorm: Should I add one here too? Seems kinda pointless otherwise since `label` already contains the same data that I'm performing this on too.

https://github.com/Authenticator-Extension/Authenticator/blob/9e0933d3e116e03c96e648e49b43f7c63c645a4b/src/background.ts#L84

Sneezry: Yeah, we should always use try catch to decode URI.

```
try {
  str = decodeURIComponent(str2)
} catch(error) {
  str = str2;
}
```

Sneezry (approved):

PR #125, opened 2018-04-18
Stop content script from executing on a tab more than once

mymindstorm:
May not be the best solution. This stops the `alert()`s from scanning a QR pile up when you try to scan the same page more than once.

### DIFF
--- a/src/content.ts
+++ b/src/content.ts
@@ -1,27 +1,28 @@
-chrome.runtime.onMessage.addListener((message, sender, sendResponse) => {
-  switch (message.action) {
-    case 'capture':
-      sendResponse('beginCapture');
-      showGrayLayout(message.passphrase);
-      break;
-    case 'errorsecret':
-      alert(chrome.i18n.getMessage('errorsecret') + message.secret);
-      break;
-    case 'errorqr':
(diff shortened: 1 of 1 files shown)


Sneezry:
A better solution is to remove the inject DOM after scan I guess

mymindstorm:
I don't think that's possible. In Firefox the only thing that gets loaded into the DOM is the `__ga_grayLayout__` div, and even if that wasn't a problem I'm not sure if the event listener would still be active.

Sneezry:
You are right, `__ga_grayLayout__` will be created again when the content script gets a message.

Another solution is checking if current tab is active when content script receives a message.

Sneezry:
Then let's check if injected:

```typescript
let injected = !!document.getElementById('__ga_grayLayout__') ||
    !!chrome.runtime.onMessage.addListener((message, sender, sendResponse) => {
  // ...
}
```

mymindstorm:
My second attempt at this involved trying to ping the content script using `sendMessage` and not inject it in the first place. Your solution is a lot less painful than trying to get myself out of my small callback hell.

Sneezry (approved):

PR #130, opened 2018-05-10
Support encrypted backups

### LINKED ISSUE #104: Export encrypted data to file
Sorry if I'm a little behind, but I just updated Authenticator to the new layout and noticed that we no longer have the ability to create a backup with encrypted secrets. We used to have that ability, and encrypted backups can still be imported.

Can we get the ability to export encrypted secrets again?

mymindstorm:
Closes #104

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ build
 chrome
 firefox
 .vscode
+.atom-build.yml
--- a/_locales/de/messages.json
+++ b/_locales/de/messages.json
@@ -76,7 +76,7 @@
         "description": "About."
     },
     "export_import": {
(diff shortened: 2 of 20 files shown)


Sneezry:
Should I approve this PR now or after all done in the to-do list?

mymindstorm:
Now, all the parts that I want you to review are already in there. I'll merge when it's ready

Sneezry:
If I provide a wrong password when I import encrypted backup file, the page will just close as normal, but no data imported.

Can we show an alert if the password is wrong? However, this is not a must-do task, it just works well, and won't break the extension.

Sneezry (approved):

PR #143, opened 2018-06-08
fix import unencrypted backup file issue, add auto paste code to input box

Sneezry:


mymindstorm:
Firefox doesn't like async permission requests:
```
Error: permissions.request may only be called from a user input handler
```

Is it a good idea to just send the code even though it won't be used? Would it be better to query if a 2fa field is there and then send the code?

Sneezry:
It should work now

src/content.ts
```diff
@@ -160,3 +163,37 @@ function showQrCode(msg: string) {
...
+  const inputBoxes: HTMLInputElement[] = [];
+  for (let i = 0; i < _inputBoxes.length; i++) {
+    if (_inputBoxes[i].type === 'text' || _inputBoxes[i].type === 'number') {
+      inputBoxes.push(_inputBoxes[i]);
+    }
+  }
+  if (!inputBoxes.length) {
+    return;
+  }
+  const identities = ['2fa', 'otp', 'authenticator', 'factor'];
```
mymindstorm: I've seen some sites use `code` too

src/content.ts
```diff
@@ -160,3 +163,37 @@ function showQrCode(msg: string) {
...
+
+  const activeInputBox = document.activeElement.tagName === 'INPUT' ?
+      document.activeElement :
+      null;
+  if (activeInputBox) {
+    (activeInputBox as HTMLInputElement).value = code;
+    return;
+  }
+  const firstInputBox = inputBoxes[0];
+  firstInputBox.value = code;
```
mymindstorm: This is destructive to data already in the input box. Either remove this section and let user paste code or check if data is already in input box. This also goes for line 193

mymindstorm (changes requested):

mymindstorm (approved):

mymindstorm:
Not exactly sure if it's ok to insert the code to a page, it seems to be against [Mozilla policy](https://developer.mozilla.org/en-US/Add-ons/AMO/Policy/Reviews#No_Surprises).

Specifically:

> Makes unexpected changes to the browser or web content

> All changes must be “opt-in”, meaning the user has to take non-default action to enact the change. Changes that prompt users via the permissions system don’t require an additional opt-in.

To be on the safe side it might be necessary to add a setting to turn this on. Not exactly sure where a setting for that would go though.

Sneezry:
Wondering how does 1Password do on Firefox. It will fill username and password on web page on Chrome by one click.

mymindstorm:
> Users should be able to easily discern what the functionality of your add-on is and not be presented with unexpected user experiences after installing it.

> “Unexpected” features are those that are unrelated to the add-on’s primary function, and are not likely from the add-on name or description to be expected by a user installing that add-on.

A password manager's browser extension's primary function is to read passwords from and insert passwords  to form fields and send them back to the main application. Authenticator's main function is to store and display 2fa codes. 

Even if you successfully argue that it's related to Authenticator's primary function, it's not made apparent anywhere that clicking on a code will insert it into the current page. 1Password and other password managers' browser extensions usually show a selector box next to password input to select an account making it less unexpected.

Sneezry:
How about just disable auto fill for Firefox currently?

mymindstorm:
All that's needed for this to be compliant is an opt-in checkbox. If you merge this as it is now, I can add a checkbox for it before release.

Sneezry:
Cool, then I merge it.

Issue #140, opened 2018-05-29
Search, sort, sync setting and more feature request


isumairu:
If you could fulfill these requests from me, that would be great.

1. Add search feature (I find Authenticator often fail to show only accounts of the site opened) and sort setting (I have many accounts and this extension seems to show them in no order. help me).
2. Add tooltips and more detailed description. What's dropbox's icon doing on the top bar? At what interval does Authenticator backup to dropbox? And other places that doesn't have explanation or hint. More infos wouldn't hurt don't you think?
3. In the export/Import setting, I suggest changing "Download" to "Save". It'd make more sense.
4. Add option to not sync to google (in chrome) and/or firefox.
5. Add option to manually start backup to dropbox.
6. Add option to launch Authenticator as a stand alone window (so it look like Authy desktop app).
7. I don't know if this is possible, but, can you make Authenticator import from dropbox or other cloud? automatically? I want synchronization between my chrome & firefox.

That's it for now. 
Thank you.

mymindstorm:
1. See #138, In regards to sorting you can go into edit mode and drag them to change their order.
2. That's a good idea, but it will take a while to translate the strings for any tooltips. The Dropbox icon appears when Dropbox sync is enabled. It attempts to sync to Dropbox every 30 days, if you need more info here is the relevant source:
https://github.com/Authenticator-Extension/Authenticator/blob/87d812deb814baee9b83a9ddefe3d140df431a6e/src/popup.ts#L54-L65
https://github.com/Authenticator-Extension/Authenticator/blob/01c972ba6a3d0f17b6ead6239833eccca07824ef/src/models/dropbox.ts#L11-L64

3. Save is a bit ambiguous, whereas if we use download you know instantly that it will save a file to your downloads folder.
4. We rely on the sync storage space, if you don't want to sync it you need to turn off extension data sync in chrome sync settings if possible.
5. That is a good feature request
6. Being a WebExtnesion, that's not really in scope of this project. If you want to make an electron fork, feel free to do so.
7. Before we work on anything like that there are some other sync issues that need to be fixed.

isumairu:
Thanks for your response.

Regarding your No. 1 reply, I'm aware of the drag-sort feature, but it's not enough nor practical for me. It can't even scroll continuously (I said I have many accounts, right?). Also, I still need search feature. Please.

Regarding No. 6, I have several question. 
[...]

PR #144, opened 2018-06-09
Enhancements

mymindstorm:
progress on #140

### DIFF
--- a/src/ui/menu.ts
+++ b/src/ui/menu.ts
@@ -1,5 +1,6 @@
 /* tslint:disable:no-reference */
 /// <reference path="../models/interface.ts" />
+/// <reference path="../models/dropbox.ts" />
 /// <reference path="./ui.ts" />
 
 function getVersion() {
@@ -125,6 +126,15 @@ async function menu(_ui: UI) {
             height: 480,
             width: 320
           });
+      },
+      dropboxUpload: async () => {
+        let dbox = new Dropbox
+        let response = await dbox.upload(_ui.instance.encryption)
+        if (response === true) {
+            _ui.instance.alert(_ui.instance.i18n.updateSuccess)
+          } else {
+            _ui.instance.alert(_ui.instance.i18n.updateFailure)
+          }
       }
     }
   };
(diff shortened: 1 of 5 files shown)


src/ui/menu.ts
```diff
@@ -117,6 +118,23 @@ async function menu(_ui: UI) {
               return;
             });
         return;
+      },
+      popOut: () => {
+        chrome.windows.create({
+          url: chrome.extension.getURL('view/popup.html'),
+          type: 'detached_panel',
```
Sneezry: Chrome doesn't support this type for extension:

`Uncaught Error: Invalid value for argument 1. Property 'type': Value must be one of: [normal, popup, panel].`

We need disable this feature on Chrome or change it to another type.

src/ui/menu.ts
```diff
@@ -117,6 +118,29 @@ async function menu(_ui: UI) {
...
+        if (navigator.userAgent.indexOf('Firefox') !== -1) {
+          windowType = 'detached_panel';
+        } else {
+          windowType = 'panel';
+        }
+        chrome.windows.create({
+          url: chrome.extension.getURL('view/popup.html'),
+          type: windowType,
+          height: 480,
+          width: 320
```
Sneezry: It seems here's a bug of Chrome, the opened window is not in width of 320px

mymindstorm: That's odd, I'm using Chromium and it seems to be fine
image

Sneezry: Indeed. It seems Chrome for windows doesn't respect that value:

snipaste20180611_102400

Sneezry (approved):

src/ui/menu.ts
Sneezry: It's not a big problem, we can fix it in the later version.

mymindstorm: Looks like if the size is updated after creating the window it might fix the issue.

I also found an old bug, but it's probably irrelevant. 
https://bugs.chromium.org/p/chromium/issues/detail?id=173831

PR #151, opened 2018-06-13
Search

### LINKED ISSUE #138: Seach feature
will be so hard add a search for this extensión?

mymindstorm:
closes #138 

Hit `/` to open search. Search bar will show if you have more than 10 accounts

I don't think the colors look that great. Do you have any better ideas on that front?

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -246,5 +246,9 @@
     "download_enc_backup": {
       "message": "Download Password-Protected Backup",
       "description": "Download Encrypted Backup"
+    },
+    "search": {
+      "message": "Search",
+      "description": "Search"
     }
 }
--- a/css/popup.css
+++ b/css/popup.css
@@ -227,6 +227,28 @@ body {
     display: none;
 }
 
+#search {
+    position: relative;
+}
+
+#searchHint {
+    position: absolute;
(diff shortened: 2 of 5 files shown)


Sneezry:
Should it be better if we hide the search bar by default?

mymindstorm:
It should be hidden by defualt if you have less than 10 accounts.

Sneezry:
screen shot 2018-06-20 at 11 21 59

It seems not on my side.

However, I find if the entries have been filtered in current tab, the search bar is hidden by default.

Sneezry (approved):

mymindstorm:
Sorry, I don't think I communicated that very well. The search bar should only show up without the user calling it when a user has more than 10 accounts. Search should not show without user intervention if the user has more than 10 accounts and auto-filter is active. Search should not show if the user has less than 10 accounts.

Before this is merged, I want to discuss the following:

- Should the minimum number of accounts for search to show by default be raised higher?
- I don't think the color for the search bar is appropriate since it shouldn't grab attention. Additionally, the hotkey hint doesn't have much contrast with the current background color.

gek15:
I

gek15:
K

gek15:
I

PR #163, opened 2018-07-11
Legal update

mymindstorm:


view/popup.html
```diff
@@ -93,16 +93,21 @@
...
-                <p><a target='_blank' href='http://blog.tinisles.com/2011/10/google-authenticator-one-time-password-algorithm-in-javascript/'>totp.js</a> Copyright its author. Licensed under the MIT License.</p>
-                <p><a target='_blank' href='http://caligatio.github.com/jsSHA/'>jsSHA</a> Copyright jsSHA authors. Licensed under the BSD License.</p>
-                <p><a target='_blank' href='http://davidshimjs.github.io/qrcodejs/'>qrcode.js</a> Copyright qrcode.js author. Licensed under the MIT License.</p>
-                <p><a target='_blank' href='https://code.google.com/p/crypto-js/'>crypto-js</a> Copyright crypto-js author. Licensed under the BSD License.</p>
-                <p><a target='_blank' href='https://www.google.com/fonts/specimen/Droid+Sans+Mono'>Droid Sans Mono</a> Copyright Steve Matteson. Licensed under the Apache License.</p>
+                <p><strong>Authenticator</strong> &copy; 2018 <a target='_blank' href='https://github.com/Authenticator-Extension'>Authenticator Extension</a>. Released under the MIT License.</p>
+                <p>Google is a trademark of Google LLC.</p>
```
Sneezry, line 97: Do we use word Google in our current code?

mymindstorm: Only on the sync clock with Google button

Sneezry: Then I think we can remove this line. What's your opinion?

mymindstorm: This line needs to be kept. "Google" is a registered trademark. Lawyers are pretty aggressive when dealing with trademarks, especially since they can be lost if they aren't defended properly.

http://tsdr.uspto.gov/#caseNumber=87786172&caseType=SERIAL_NO&searchType=statusSearch

Sneezry: Emmm, okay. Let's keep it.

Sneezry (approved):

PR #169, opened 2018-07-25
Fix popup size issue

mymindstorm:
This _should_ fix the popup size issue, but I don't have a windows computer to test with right now.

### DIFF
--- a/src/ui/menu.ts
+++ b/src/ui/menu.ts
@@ -145,6 +145,24 @@ async function menu(_ui: UI) {
             new URLSearchParams(document.location.search.substring(1));
         return params.get('popup');
       },
+      fixPopupSize: () => {
+        const zoom = Number(localStorage.zoom) / 100 || 1;
+        const correctHeight = 480 * zoom;
+        const correctWidth = 320 * zoom;
+        if (window.innerHeight !== correctHeight ||
+            window.innerWidth !== correctWidth) {
+          chrome.windows.getCurrent((currentWindow) => {
+            // window update to correct size
+            const adjustedHeight =
+                window.innerHeight + (correctHeight - window.innerHeight);
+            const adjustedWidth =
+                window.innerWidth + (correctWidth - window.innerWidth);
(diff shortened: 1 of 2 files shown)


src/ui/menu.ts
```diff
@@ -145,6 +145,24 @@ async function menu(_ui: UI) {
...
+      fixPopupSize: () => {
+        const zoom = Number(localStorage.zoom) / 100 || 1;
+        const correctHeight = 480 * zoom;
+        const correctWidth = 320 * zoom;
+        if (window.innerHeight !== correctHeight ||
+            window.innerWidth !== correctWidth) {
+          chrome.windows.getCurrent((currentWindow) => {
+            // window update to correct size
+            const adjustedHeight =
+                window.innerHeight + (correctHeight - window.innerHeight);
```
Sneezry: doesn't it equal to `correctHeight`?

mymindstorm: I'm trying to account for the window frame here. I think the issue with the height stems from windows.create needing us to account for the window frame in Windows, whereas in Linux you don't need to. So if there is an offset needed for the window frame this should add whatever that is to `correctHeight`

Sneezry: I think `chrome.windows.update` is the key, not `adjustedHeight`.

try
```
chrome.windows.update(
  currentWindow.id,
  {height: correctHeight, width: correctWidth});
});
```

I guess it should work.

mymindstorm: At least windows.create is supposed to take the frame into account according to [this](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/windows/create#Paramaters), it's pretty silly for windows.update to behave differently if that is the case.

Sneezry: Have no idea if this issue is related with this bug: https://bugs.chromium.org/p/chromium/issues/detail?id=173831

This bug is still not marked as resolved.

Sneezry (approved):

PR #175, opened 2018-07-29
Port to edge

mymindstorm:
I thought edge was ok before, but now I loathe it. 

Sideloading unpacked extensions (not sideloading an appx package) will cause it to have weird issues that don't happen when loaded via `Add-AppxPackage`. Using `Add-AppxPackage` is great except for that fact that storage.sync stops working altogether.

That type of issue is manageable at least, unlike this lovely utility. `Add-AppxPackage` at first glance is a great utility. Sadly, it turs out that `Add-AppxPackage` is stupid. I can't use a packed extension without signing it (understandable). But if you aren't careful when loading an unpacked appx this happens:
image

Lo and behold, I need permission from _myself_ to delete this folder. I'm probabbly going to have to go to safe mode just to delete a folder. The constant crashing when using debugging tools isn't really helping either.

---

In terms of actual problems, the port works for the most part. You can't download a backup, it's a [known issue](https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/14603958/). Links to extension pages don't work either (E.g. "Import Backup"), still have to experiment with that. All I could find were old issues either [resolved](https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/7191389/) or marked as [not reproducible](https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/8150728/). I've made an issue in the public bug tracker for now. Everything else seems to work perfectly (except dropbox, identity isn't supported).

You can use `installEdge` or `packageEdge` to install it. All the scripts I added work with [Cygwin](http://www.cygwin.com/) (C: is /cygdrive/c). `packageEdge` uses the makeappx.exe that the App Certification Kit installs, so yours might be in a different location and cause the packaging script to fail.


Sneezry:
Interesting. Does Edge support `chrome` namespace?

mymindstorm:
There is a tool called the Edge Extension Toolkit that makes those ABIBridge scripts. I think those provide support for the chrome namespace.

mymindstorm:
It's confirmed now. https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/18438862

mymindstorm:
Ok so DO NOT under any circumstances try to run the build script again until you fully uninstall the extension from edge or the folder thing happens.

mymindstorm:
That's everything! I've tested all of it, so if the code looks fine we can merge this.

Sneezry (approved):

PR #189, opened 2018-08-11
Add BrowserStorage class and fix the popup size issue for the last time

mymindstorm:
This lets me kill off the `edge-tempfix` branch and, in the future, will allow us to give the user the choice to store data locally instead of always in `sync`. Still need to make a settings UI and migration function for that. Please scrutinize this since mistakes in this would cause lots of trouble.

Not exactly sure why 350-480 got changed. New to VS Code, maybe my config is wrong?

### DIFF
--- a/src/models/storage.ts
+++ b/src/models/storage.ts
@@ -8,21 +8,21 @@ class BrowserStorage {
     if (localStorage.storageLocation !== 'sync' &&
         localStorage.storageLocation !== 'local') {
       return new Promise((resolve, reject) => {
-        let sync: number;
-        let local: number;
-        chrome.storage.local.get((l) => {
-          local = Object.keys(l).length;
+        let amountSync: number;
+        let amountLocal: number;
+        chrome.storage.local.get((local) => {
+          amountLocal = Object.keys(local).length;
           try {
-            chrome.storage.sync.get((s) => {
-              sync = Object.keys(s).length;
+            chrome.storage.sync.get((sync) => {
+              amountSync = Object.keys(sync).length;
               // If storage location can't be found try to auto-detect storage
               // location
-              if (local > sync && sync === 0) {
+              if (amountLocal > amountSync && amountSync === 0) {
                 localStorage.storageLocation = 'local';
-              } else if (local < sync && local === 0) {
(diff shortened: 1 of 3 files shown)


Sneezry:
Is there a way to set `localStorage.storageLocation` value? What if I save all data in sync but change to local then? Will the data be migrated to the other scope?

mymindstorm:
`localStorage.storageLocation` is currently set by this promise if it does not exist https://github.com/Authenticator-Extension/Authenticator/blob/d039de55eef58562cb17c4ae431dd71d85287519/src/models/storage.ts#L10-L38

Currently, the only reason this exists is to merge the codebases together and to lay the groundwork for letting the user change their storage space. There is no way for the user to change the storage space yet, hence there is no migration code. That's going to be my next PR.

src/models/storage.ts
```diff
@@ -3,6 +3,79 @@
...
 /// <reference path="./otp.ts" />
 
+class BrowserStorage {
+  private static getStorageLocation() {
+    if (localStorage.storageLocation !== 'sync' &&
+        localStorage.storageLocation !== 'local') {
+      return new Promise((resolve, reject) => {
+        let sync: number;
+        let local: number;
+        chrome.storage.local.get((l) => {
```
Sneezry: Better naming for `l`

src/models/storage.ts
```diff
@@ -3,6 +3,79 @@
...
+          local = Object.keys(l).length;
```
Sneezry: Better naming for `local`. This should be something like number of keys store in local, right?

src/models/storage.ts
```diff
@@ -3,6 +3,79 @@
...
+                localStorage.storageLocation = 'sync';
+              } else {
+                // No or duplicate user data, use defualt
+                if (navigator.userAgent.indexOf('Edge') !== -1) {
+                  localStorage.storageLocation = 'local';
+                } else {
+                  localStorage.storageLocation = 'sync';
+                }
+              }
+              resolve(localStorage.storageLocation);
```
Sneezry, line 32: It is better always return after resolve or reject in a promise. `return resolve(...)` or `resolve(...); return;`

Sneezry (approved):

PR #192, opened 2018-08-25
Storage migrate code

mymindstorm:


src/ui/menu.ts
```diff
@@ -75,7 +75,12 @@ async function menu(_ui: UI) {
   let useAutofill = (localStorage.autofill === 'true');
 
   const ui: UIConfig = {
-    data: {version, zoom, useAutofill},
+    data: {
+      version,
+      zoom,
+      useAutofill,
+      newStorageLocation: localStorage.storageLocation
```
Sneezry, line 82: `newStorageLocation` should have a default value. `newStorageLocation: localStorage.storageLocation || 'sync'`

mymindstorm: Shouldn't default be null? Worst case scenario the storage space select box is blank on first run, that's better than it being potentially incorrect.

Sneezry:
I see. Or something should be shown insread of blank.

mymindstorm:
Is this ok as it is? It won't ever show up as blank after 2b4d26083d85f29098330bd94f3146f35dd9ceda as long as [this](https://github.com/Authenticator-Extension/Authenticator/blob/6ba9c476b11d1d94ec0ae59f6aa3bb26df53c562/src/models/storage.ts#L35) doesn't trigger.

Sneezry (approved):

Sneezry:
ok, lets check it in

PR #195, opened 2018-09-08
Google Drive support

mymindstorm:
This is still a work in progress.

- I can't get `identity.launchWebAuthFlow` to work properly with anything I compile on Chrome, even with known working versions. There are likely still kinks in Chrome since I haven't been able to test. Still haven't figured out why it's giving me trouble.

Todo:
- [x] Still need to have it dump everything in a folder and not the root directory
- [x] Consolidate code
- [x] 40bac1adea18765554910757d9098a940b0fe07c

### DIFF
--- a/_locales/de/messages.json
+++ b/_locales/de/messages.json
@@ -76,7 +76,7 @@
         "description": "About."
     },
     "export_import": {
-        "message": "Exportieren \/ Importieren",
+        "message": "Exportieren / Importieren",
         "description": "Export and Import."
     },
     "settings": {
@@ -290,5 +290,13 @@
     "storage_location": {
         "message": "Storage Location",
         "description": "Storage location"
+    },
+    "sign_in": {
+        "message": "Sign in",
+        "description": "Sign in to 3rd party storage services"
+    },
+    "log_out": {
+        "message": "Sign out",
+        "description": "Sign out of 3rd party storage services"
     }
 }
--- a/_locales/de/messages.json
+++ b/_locales/de/messages.json
@@ -296,7 +296,7 @@
         "description": "Sign in to 3rd party storage services"
     },
     "log_out": {
-        "message": "Sign out",
+        "message": "Logout",
         "description": "Sign out of 3rd party storage services"
     }
 }
--- a/src/background.ts
+++ b/src/background.ts
@@ -3,6 +3,7 @@
 /// <reference path="./models/encryption.ts" />
 /// <reference path="./models/interface.ts" />
 /// <reference path="./models/storage.ts" />
+/// <reference path="./models/credentials.ts" />
 
 let cachedPassphrase = '';
 
@@ -158,11 +159,14 @@ function getBackupToken(service: string) {
   let authUrl = '';
   if (service === 'dropbox') {
     authUrl =
-        'https://www.dropbox.com/oauth2/authorize?response_type=token&client_id=mmx38seexw3tvps&redirect_uri=' +
-        encodeURIComponent(chrome.identity.getRedirectURL());
+        'https://www.dropbox.com/oauth2/authorize?response_type=token&client_id=' +
+        getCredentials().dropbox.client_id +
+        '&redirect_uri=' + encodeURIComponent(chrome.identity.getRedirectURL());
   } else if (service === 'drive') {
     authUrl =
-        'https://accounts.google.com/o/oauth2/v2/auth?response_type=code&access_type=offline&client_id=292457304165-ria4acohb2i875o1kmda5a31vkan7rj7.apps.googleusercontent.com&scope=https%3A//www.googleapis.com/auth/drive.file&prompt=consent&redirect_uri=' +
+        'https://accounts.google.com/o/oauth2/v2/auth?response_type=code&access_type=offline&client_id=' +
+        getCredentials().drive.client_id +
+        '&scope=https%3A//www.googleapis.com/auth/drive.file&prompt=consent&redirect_uri=' +
         encodeURIComponent('https://authenticator.cc/oauth');
   }
   chrome.identity.launchWebAuthFlow({url: authUrl, interactive: true}, async (url) => {
@@ -205,8 +209,10 @@ function getBackupToken(service: string) {
                  reject: (reason: Error) => void) => {
                   xhr.open(
                       'POST',
-                      'https://www.googleapis.com/oauth2/v4/token?client_id=292457304165-ria4acohb2i875o1kmda5a31vkan7rj7.apps.googleusercontent.com&client_secret=0AQA9EDq-WHkPLX4mfcIZ4ws&code=' +
-                          value +
+                      'https://www.googleapis.com/oauth2/v4/token?client_id=' +
+                          getCredentials().drive.client_id + '&client_secret=' +
+                          getCredentials().drive.client_secret +
+                          '&code=' + value +
                           '&redirect_uri=https://authenticator.cc/oauth&grant_type=authorization_code');
                   xhr.setRequestHeader('Accept', 'application/json');
                   xhr.setRequestHeader(
--- a/src/background.ts
+++ b/src/background.ts
@@ -18,10 +18,8 @@ chrome.runtime.onMessage.addListener((message, sender, sendResponse) => {
     cachedPassphrase = message.value;
   } else if (message.action === 'passphrase') {
     sendResponse(cachedPassphrase);
-  } else if (message.action === 'dropbox') {
-    getDropboxToken();
-  } else if (message.action === 'drive') {
-    getDriveToken();
+  } else if (['dropbox', 'drive'].indexOf(message.action) > -1) {
+    getBackupToken(message.action);
   } else if (message.action === 'lock') {
     cachedPassphrase = '';
   }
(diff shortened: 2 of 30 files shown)


mymindstorm:
@Sneezry this is ready for review

_locales/de/messages.json
```diff
@@ -290,5 +290,13 @@
     "storage_location": {
         "message": "Storage Location",
         "description": "Storage location"
+    },
+    "sign_in": {
+        "message": "Sign in",
+        "description": "Sign in to 3rd party storage services"
+    },
+    "log_out": {
+        "message": "Sign out",
```
Sneezry: Logout is better here to distinguish from sign in

src/background.ts
```diff
@@ -154,15 +154,19 @@ async function getTotp(text: string, passphrase: string) {
...
+function getBackupToken(service: string) {
+  let authUrl = '';
+  if (service === 'dropbox') {
+    authUrl =
+        'https://www.dropbox.com/oauth2/authorize?response_type=token&client_id=mmx38seexw3tvps&redirect_uri=' +
+        encodeURIComponent(chrome.identity.getRedirectURL());
+  } else if (service === 'drive') {
+    authUrl =
+        'https://accounts.google.com/o/oauth2/v2/auth?response_type=token&client_id=292457304165-ria4acohb2i875o1kmda5a31vkan7rj7.apps.googleusercontent.com&scope=https%3A//www.googleapis.com/auth/drive.file&prompt=consent&redirect_uri=' +
+        encodeURIComponent('https://authenticatortest.tk');
```
Sneezry: Is this uri available? If callback uri is required, shall we use authenticator.cc instead?

src/models/backup.ts
```diff
@@ -64,3 +64,160 @@ class Dropbox {
...
+    }
+    return localStorage.driveFolder;
+  }
+
+  async upload(encryption: Encryption) {
+    if (localStorage.driveEncrypted === undefined) {
+      localStorage.driveEncrypted = 'true';
+    }
+    const exportData = await EntryStorage.getExport(
+        encryption, (localStorage.driveEncrypted === 'true') ? true : false);
```
Sneezry: `localStorage.driveEncrypted === 'true'` is the same as `(localStorage.driveEncrypted === 'true') ? true : false`

src/background.ts
mymindstorm: Yes, I have control over this url. Will make a blank page on authenticator.cc for redirect.

mymindstorm:
Looks like this isn't quite done yet, oops.
- [x] Token expires in an hour. See: [here](https://developers.google.com/identity/protocols/OAuth2WebServer#offline)
- [x] `backupLogout` should call [Dropbox](https://www.dropbox.com/developers/documentation/http/documentation#auth-token-revoke) and [Google](https://developers.google.com/identity/protocols/OAuth2WebServer#tokenrevoke) revoke endpoints

According to Google policy:

> Verify your application’s homepage URL domain ownership with Google through Search Console by using an account that is either a Project Owner or a Project Editor on your OAuth Project.

> Ensure the top level domains for all your URLs such as privacy policy URL, redirect URI, homepage URL should match with the verified domain that you own.

@Sneezry send me the email of the account that verified authenticator.cc and I'll add to IAM for the project

Sneezry:
It should be sneezry.lee#gmail.com

mymindstorm:
Invited, thanks.

mymindstorm:
[This](https://authenticator.cc/oauth/drive-privacy.html) is also required by Google policy. After you look at that and the new commits this will finally be over.

I really should have done OneDrive first, the Microsoft Graph documentation is much nicer than Google's.

Sneezry:
Cool!

Sneezry (approved):

mymindstorm:
I'm exposing our Google client secret in the source, is that ok? 

https://tools.ietf.org/html/rfc6819#section-4.1.1

To my understanding the client secret is used to verify if the client is trusted. For us that's pretty much impossible since any other webextension could easily impersonate us. Thankfully, chrome and firefox are pretty strict about unsigned extensions.

In the event that an attacker gets their hands on our client tokens, I doubt that they would really care since I'm using a scope that doesn't relinquish any data other than what would already be given up by a compromised client.

tdr-its42:
Hi, Maybe you should read this about exposing your client secret: https://stackoverflow.com/questions/27585412/can-i-really-not-ship-open-source-with-client-id/28109307#28109307
However, I don't know if this still a valid thread as it's already 3 years old...

mymindstorm:
@tomderidder thanks for that, didn't come up during my research. 

https://developers.google.com/terms/#b_confidential_matters

> Developer credentials (such as passwords, keys, and client IDs) are intended to be used by you and identify your API Client. You will keep your credentials confidential and make reasonable efforts to prevent and discourage other API Clients from using your credentials. Developer credentials may not be embedded in open source projects. 

What's funny is that Google exposes their own secrets in [their example code](https://github.com/googlesamples/oauth-apps-for-windows/blob/a9c06c539adc21eab752537b234219c448356001/OAuthDesktopApp/OAuthDesktopApp/MainWindow.xaml.cs#L35-L36).

Also: https://github.com/search?q=.apps.googleusercontent.com+secret&type=Code

Sneezry:
I haven't made a complete search, but as far as I know, secret is not needed for oauth if we use App Id and redirect URL.

Redirect URL is enough to prove who we are. Here's the message I talked with Dropbox team about oauth:

> Lee Sneezry
> March 13, 2018 13:14
> Hi,
> 
> My request for product was rejected (App Key: 013qun2m82h9jim, App Name: Authenticator Extension), because I put the secret key in the source code, which is public.
> 
> I agree that public secret key does not make sense. However, my question is, do I have to host a server to request the token for user, or there is another way to do it correctly?
> 
> I found this page about oauth guide: https://www.dropbox.com/developers/reference/oauth-guide, but the picture does not clarify how the app talk to dropbox after user allows the request.
> 
> Appreciate any help, thanks!
> 
> Regards, 
> Zhe

---

> Gregory
> March 13, 2018 23:50
> Hi Lee,
> 
> Thanks for writing in. You do not need a server to handle the OAuth app authorization flow. Depending on your app/environment and how you wish to implement it, there are a few options for handling the flow without your own server:
> 
> First, you should decide if you should use the "token", or "code" flow. The token flow is meant for client-side apps. The code flow is meant for server-side apps. You can find more information on both in the documentation:
> 
> https://www.dropbox.com/developers/documentation/http/documentation#oauth2-authorize
> 
> If you use the token flow, a redirect URI is required. That just needs to be some URI that your app can see; it doesn't need to be an actual remote server (e.g., it can be a local address). When using the token flow, the app secret is not even needed, so it can be left out of your source/app.
> 
> If you use the code flow, a redirect URI is not actually required. If omitted, the user will be presented with an authorization code that they can manually copy/paste into the app. The app does need the app secret to complete the flow when using the code flow though.
> 
> Regards,
> Gregory

---

> Lee Sneezry
> March 14, 2018 00:08
> Hi Gregory,
> 
> Thanks for reply, your answer is really helpful!
> 
> As you mentioned, the token flow needs a redirect URL, which can be a local address, this method inspires me. What I am developing is a Chrome extension, and I really have local address. However, page in Chrome extension is with address starts with chrome-extension://, but not http:// or https://, is it okay for the token flow? As I know, it seems I cannot set redirect URL starts with non-http or non-https.
> 
> Chrome extension doesn't have ability to start a HTTP server in the browser (Chrome app does, but it cannot access to the browser).
> 
> Regards, 
> Zhe

---

> Gregory
> March 14, 2018 00:14
> Hi Lee,
> 
> I believe you should be able to use a custom URI, such as chrome-extension://, with the token flow. Can you share whatever error you're getting when attempting to do so?
> 
> Alternatively, if your extension has access to https:// URLs, you may want to use use one of these redirect URIs that Dropbox provides for use with the OAuth flow:
> 
> This one doesn't show anything to the user, and pops the information off the URL (but your app should grab it first): https://www.dropbox.com/1/oauth2/redirect_receiver 
> This one displays the access token to the user: https://www.dropbox.com/1/oauth2/display_token
> 
> Regards,
> Gregory

---

> Lee Sneezry
> March 14, 2018 02:22
> Oh, my bad. I just thought redirect URL cannot start with chrome-extension://, because placeholder in the input box is https://. I tested to add chrome-extension:// URL, and I can set it in app settings. I will try token flow in this way, thanks!
> 
> This ticket can be closed, if I have other question about redirect URL, I will open a new one.
> 
> Thanks, Gregory! You helped me very much!
> 
> Regards, 
> Zhe

mymindstorm:
Google auth is a bit different from Dropbox. Implicit grant ("token" flow) tokens will only last for an hour, whereas Dropbox tokens practically last forever. Because of that we have to use an authorization code grant ("code" flow) which is generally used for server side apps, but it looks like it can be used for native apps in some cases.

https://developers.google.com/identity/protocols/OAuth2#installed

>  The Google OAuth 2.0 endpoint supports applications that are installed on devices such as computers, mobile devices, and tablets. When you create a client ID through the Google API Console, specify that this is an Installed application, then select Android, Chrome, iOS, or "Other" as the application type.
>
> The process results in a client ID and, in some cases, a client secret, which you embed in the source code of your application. (In this context, the client secret is obviously not treated as a secret.) 

https://developers.google.com/identity/protocols/OAuth2InstalledApp#offline

> Note: Due to the fact that the client cannot keep the client_secret confidential, you cannot do incremental authorization with installed apps.

> The client secret obtained from the API Console. (The client_secret is not applicable to requests from clients registered as Android, iOS, or Chrome applications.) 

The Chrome type only works if you use [getAuthToken](https://developer.chrome.com/apps/identity#method-getAuthToken) which is Google-specific and not supported in browsers other than Chrome. The iOS type uses a format of redirect URI that can't be parsed by launchWebAuthFlow. Sending a request without our `client_secret` results in an error. Overall, I think this as it is now should be fine since we aren't violating the terms. If we get rid of the client secret we will still have to manually put in the client id thanks to Google's tos. There isn't much of a security issue since the client secret is only used to generate access tokens from the refresh token.

mymindstorm:
Additionally, I don't really think hosting a proxy server that does token requests would make this any more secure / solve anything. There is no definite way for us to tell if requests come from a client or attacker. All that would really do is switch the target from Google directly to us.

Sneezry:
If there is no better way to handle oauth for Google Drive, let's just commit the secret.

mymindstorm:
Are you sure? That would put us in violation of the Google APIs Terms of Service.

Sneezry:
As you mentioned above, Google itself commits its secret. Also, from Google official docs, Chrome is listed as installed application scenario, which uses App ID and secret for oauth. 

> Installed applications
> 
> The Google OAuth 2.0 endpoint supports applications that are installed on devices such as computers, mobile devices, and tablets. When you create a client ID through the Google API Console, specify that this is an Installed application, then select Android, Chrome, iOS, or "Other" as the application type.
> 
> The process results in a client ID and, in some cases, a client secret, which you embed in the source code of your application. (In this context, the client secret is obviously not treated as a secret.)

mymindstorm:
I added a little warning to the build script for forks.

> The client_secret is not applicable to requests from clients registered as Android, iOS, or Chrome applications.

PR #226, opened 2018-10-14
Show warning when auth fails

mymindstorm:


src/models/backup.ts
```diff
@@ -157,6 +166,11 @@ class Drive {
 
   private async getFolder() {
     const token = await this.getToken();
+    if (!token) {
+      return new Promise((resolve: (value: false) => void) => {
```
Sneezry: do you mean `value: boolean` ?

src/models/backup.ts
```diff
@@ -250,6 +264,11 @@ class Drive {
     const backup = JSON.stringify(exportData, null, 2);
 
     const token = await this.getToken();
+    if (!token) {
+      return new Promise((resolve: (value: false) => void) => {
```
Sneezry: same here

src/models/backup.ts
```diff
@@ -132,11 +136,16 @@ class Drive {
...
                   localStorage.removeItem('driveRefreshToken');
+                  localStorage.driveRevoked = true;
                   resolve(false);
                 }
                 try {
                   const res = JSON.parse(xhr.responseText);
                   if (res.error) {
+                    if (res.error === 'invalid_grant') {
+                      localStorage.removeItem('driveRefreshToken');
+                      localStorage.driveRevoked = true;
```
Sneezry, line 147: do we need to clear token cache here?

mymindstorm: No, that only applies to getAuthToken (chrome only). Google doesn't give a 401 when the token gets revoked, and it wasn't actually deleteing them properly before.

src/models/backup.ts
mymindstorm: Since it is only supposed to return false is it ok to just leave it like this or is it better to use Boolean?

Sneezry: `false` is not a type, `boolean` should be used here.

Sneezry (approved):

bunchuairak:
eobot.com

ในวันที่ อ. 16 ต.ค. 2018 เวลา 09:38 Zhe Li <notifications@github.com>
เขียนว่า:

PR #252, opened 2018-12-04
fix chrome drive sync issues

mymindstorm:
- Fixes issue with text not being substituted
- Fixes issue where failure message shows up instead of auth_revoke message


src/models/backup.ts
```diff
@@ -83,7 +83,9 @@ class Drive {
                     const res = JSON.parse(xhr.responseText);
                     if (res.error) {
                       if (res.error.code === 401) {
-                        localStorage.driveToken = '';
+                        if (navigator.userAgent.indexOf('Chrome') === -1) {
```
Sneezry: Any detail about the exception for Chrome? Why should we not remove cache in Chrome?

src/ui/backup.ts
```diff
@@ -19,7 +19,7 @@ async function backup(_ui: UI) {
             _ui.instance.alert(_ui.instance.i18n.updateSuccess);
           } else if (localStorage.dropboxRevoked === 'true') {
             _ui.instance.alert(String(_ui.instance.i18n.token_revoked)
-                                   .replace(/\${service}/, 'Dropbox'));
+                                   .replace(/\$?{?service}/, 'Dropbox'));
```
Sneezry: It seems not make sense. `/\$?{?service}/` matches `${service}`, `{service}`, `$service}` and `service}`

src/models/backup.ts
mymindstorm: It's so it can remove the token from the identity cache, and after that remove the token from storage. Otherwise it will just keep returning the revoked token.
https://github.com/Authenticator-Extension/Authenticator/blob/777d47a2751336948b957a84f5ccf975ca786103/src/models/backup.ts#L111-L114

src/ui/backup.ts
mymindstorm: The most I can do to make it more specific is this: `/(?:\${)?service}/`.

Why:
Firefox
image
Chrome
image

src/models/backup.ts
Sneezry: Then why should we skip clearing cache in Chrome

src/ui/backup.ts
Sneezry: `${variable}` is parsed as variable reference in JS. `{{replacement}}` would be better for template use.

Sneezry: It seem i18n message supports replacement. See https://developer.chrome.com/extensions/i18n#examples-getMessage
and
https://chromium.googlesource.com/chromium/src/+/master/chrome/common/extensions/docs/examples/extensions/news/_locales/en/messages.json#73

src/models/backup.ts
mymindstorm: You can't remove a token from the cache without knowing what it is. After the chrome cached token is removed, the token is removed from localStorage like normal.

kandikane (approved):

src/models/backup.ts
Sneezry: I still have no idea why we need to check if the current runtime is not Chrome. Can you add a comment in the code here?

mymindstorm: Does cd95c06 make more sense?

mymindstorm:
@Sneezry is this okay to merge?

src/models/backup.ts
```diff
@@ -77,12 +77,18 @@ class Drive {
...
               xhr.setRequestHeader(
                   'Authorization', 'Bearer ' + localStorage.driveToken);
-              xhr.onreadystatechange = () => {
+              xhr.onreadystatechange = async () => {
                 if (xhr.readyState === 4) {
                   try {
                     const res = JSON.parse(xhr.responseText);
                     if (res.error) {
                       if (res.error.code === 401) {
+                        if (navigator.userAgent.indexOf('Chrome') !== -1) {
```
Sneezry, line 86: can this line be removed?

mymindstorm: No. We have to handle Chrome Google account login differently (via `chrome.identity`) from Firefox because of Google API restrictions. `chrome.identity` isn't part of Firefox, hence the check here.

When a user revokes a token from [Google's side](https://myaccount.google.com/permissions) the token is still cached in Chrome. So, if a user tries to sign in again directly after revoking a token `chrome.identity.getAuthToken` would notice a cached token is still available and pull it even though it was revoked. This causes the extension to show a 'Failure' message since the 'new' token is revoked already.

To reproduce with 5.2.2:

1. login to google drive
2. sync
3. revoke access from https://myaccount.google.com/permissions
4. keep syncing until the extension gets a 401
5. close and open the extension popup (not chrome)
6. try to login again
7. get error
8. goto 6

To prove it's the cache after reproducing:

1. see error after trying to login
2. open chrome://identity-internals/ and revoke token
3. try to login to drive sync
4. success

Sneezry (approved):

PR #275, opened 2019-02-02
fix security issue

### LINKED ISSUE #272: Password-protected backups and stored settings may have unencrypted secrets
**Platform:** Linux
 - Browser: Firefox
 - Browser Version: 60.4.0esr

When trying to export the backup of my 3 tokens one ends up unencrypted!
Looking into authenticator.json one token contains 
>    "encrypted": false,
and the secret is indeed not encrypted.

I experimented and added / deleted / re-added test-tokens and ended up with various results.

It seems to forget the password sometimes, but why then only some tokens get stored unencrypted is unclear to me. Anyway this is seems pretty unsafe to me. 

If I enter the password (and verification) once I would expect to need it for every single token and would expect nothing to be stored in cleartext nowhere from then on - until I remove the password.

mymindstorm:
fixes #272

### DIFF
--- a/src/background.ts
+++ b/src/background.ts
@@ -14,7 +14,7 @@ chrome.runtime.onMessage.addListener((message, sender, sendResponse) => {
     }
     getQr(
         sender.tab, message.info.left, message.info.top, message.info.width,
-        message.info.height, message.info.windowWidth, message.info.passphrase);
+        message.info.height, message.info.windowWidth);
   } else if (message.action === 'cachePassphrase') {
     cachedPassphrase = message.value;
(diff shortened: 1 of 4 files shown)


Benno-K:
@Sneezry @mymindstorm 
Is there anything where I could help to make progress on this topic?
I do not want to put pressure on you, but on the other hand as long as the fix isn't published as - or along with - a new version, scanned secrets are quite likely to be stored unencrypted for a lot of users. 
(Off course, I myself am already using the modified code, so my secrets are safe)

Sneezry (approved):

Issue #282, opened 2019-02-16
Standardize backup format


fortran77:
We seem to have more and more open source TOTP applications. Each one can export to json but uses its own format. I think one of the developers should publish a proposed a json export format and invite everybody to use that. This would be for data exchange, so encryption is not strictly needed (though desirable).

So far we have:

[Authenticator-Extension/Authenticator](https://github.com/Authenticator-Extension/Authenticator) (this one). Relevant previous issue #185.

[Gnome Authenticator](https://gitlab.gnome.org/World/Authenticator). (Gnome desktop.) Relevant issue: [https://gitlab.gnome.org/World/Authenticator/issues/111](https://gitlab.gnome.org/World/Authenticator/issues/111).

[FreeOTP+](https://github.com/helloworld1/FreeOTPPlus) (Android app.) See above issue.

[AndOTP](https://github.com/andOTP/andOTP) (Android app.) Relevant issue: https://github.com/andOTP/andOTP/issues/255

mymindstorm:
I'm fine with doing this, but it's pointless to make a standard backup schema without any other developers on board.

See also: https://github.com/google/google-authenticator/wiki/Key-Uri-Format

fortran77:
So if a URL format of the type otpauth could be standardized, then maybe an export format could be just a series of these URLs, newline-separated. If so, then we need to get everybody to agree on the URL format.

mymindstorm:
@fortran77 that's already standardized, everyone uses it for adding individual secrets via QR codes. You wanted a standardized backup format which would hold multiple secrets.

fortran77:
Well, that's great, then! So we just need a file containing one otpauth URL per line. That would work, wouldn't it? No encryption, but all we're looking for is data exchange. If we can get just one developer to use this format for import and export, it should be easy to get everyone else to do so.

mymindstorm:
@fortran77 it would be simpler to just use JSON. It's a great idea to standardize the format, but it's pointless to do without anyone else willing to implement it. If you can find another developer willing to work with us on this then this could happen.

fortran77:
Requested that the Gnome Authenticator developers consider the one otpauth URL per line format:

[https://gitlab.gnome.org/World/Authenticator/issues/111#note_439185](https://gitlab.gnome.org/World/Authenticator/issues/111#note_439185)

mymindstorm:
[...]

PR #286, opened 2019-02-19
add otpauth url import export support

Sneezry:
https://github.com/Authenticator-Extension/Authenticator/issues/282

### DIFF
--- a/src/ui/entry.ts
+++ b/src/ui/entry.ts
@@ -67,27 +67,27 @@ function getOneLineOtpBackupFile(entryData: {[hash: string]: OTPStorage}) {
   const otpAuthLines: string[] = [];
   for (const hash of Object.keys(entryData)) {
     const otpStorage = entryData[hash];
-    let otpAuthLine = 'otpauth://';
+    const label = otpStorage.issuer ?
+        (otpStorage.issuer + ':' + otpStorage.account) :
+        otpStorage.account;
+    let type = '';
     if (otpStorage.type === 'totp' || otpStorage.type === 'hex') {
-      otpAuthLine += 'totp/';
+      type = 'totp';
     } else if (otpStorage.type === 'hotp' || otpStorage.type === 'hhex') {
-      otpAuthLine += 'hotp/';
+      type = 'hotp';
     } else {
       continue;
     }
 
-    otpAuthLine += encodeURIComponent(otpStorage.issuer || ' ') + ':' +
-        encodeURIComponent(otpStorage.account || ' ') + '?';
-    otpAuthLine += 'secret=' + otpStorage.secret;
-    otpAuthLine += '&issuer=' + encodeURIComponent(otpStorage.issuer || ' ');
+    const otpAuthLine = 'otpauth://' + type + '/' + label +
+        '?secret=' + otpStorage.secret +
+        (otpStorage.issuer ? ('&issuer=' + otpStorage.issuer.split('::')[0]) :
+                             '') +
+        (type === 'hotp' ? ('&counter=' + otpStorage.counter) : '') +
+        (type === 'totp' && otpStorage.period ?
+             ('&period=' + otpStorage.period) :
+             '');
 
-    if (otpStorage.type === 'totp' || otpStorage.type === 'hex') {
-      otpAuthLine += '&period=' + (otpStorage.period || 30);
-    }
-
-    if (otpStorage.type === 'hotp' || otpStorage.type === 'hhex') {
-      otpAuthLine += '&counter=' + otpStorage.counter;
-    }
     otpAuthLines.push(otpAuthLine);
   }
 
@@ -228,7 +228,7 @@ async function getCachedPassphrase() {
       });
 }
 
-function getExportDataFromOTPAuthPerLine(importCode: string) {
+function getEntryDataFromOTPAuthPerLine(importCode: string) {
   const lines = importCode.split('\n');
   const exportData: {[hash: string]: OTPStorage} = {};
   for (let item of lines) {
@@ -459,7 +459,7 @@ async function entry(_ui: UI) {
 
         } catch (error) {
           // Maybe one-otpauth-per line text
-          exportData = getExportDataFromOTPAuthPerLine(_ui.instance.importCode);
+          exportData = getEntryDataFromOTPAuthPerLine(_ui.instance.importCode);
         }
 
         try {
@@ -542,8 +542,14 @@ async function entry(_ui: UI) {
           const reader = new FileReader();
           let decryptedFileData: {[hash: string]: OTPStorage} = {};
           reader.onload = async () => {
-            const importData: {[hash: string]: OTPStorage} =
-                JSON.parse(reader.result as string);
+            let importData: {[hash: string]: OTPStorage} = {};
+            try {
+              importData = JSON.parse(reader.result as string);
+            } catch (e) {
+              importData =
+                  getEntryDataFromOTPAuthPerLine(reader.result as string);
+            }
+
             let encrypted = false;
             for (const hash in importData) {
               if (importData[hash].encrypted) {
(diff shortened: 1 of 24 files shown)


src/ui/entry.ts
```diff
@@ -63,6 +63,39 @@ function getBackupFile(entryData: {[hash: string]: OTPStorage}) {
   return `data:application/octet-stream;base64,${base64Data}`;
 }
 
+function getOneLineOtpBackupFile(entryData: {[hash: string]: OTPStorage}) {
```
mymindstorm, line 70: - Empty issuer should just exclude the `&issuer=` field altogether.
- Empty issuer and account field should actually be empty, not a space (`%20`) if blank
- If period is default, just leave it out

src/ui/entry.ts
```diff
@@ -63,6 +63,39 @@ function getBackupFile(entryData: {[hash: string]: OTPStorage}) {
...
+  const otpAuthLines: string[] = [];
+  for (const hash of Object.keys(entryData)) {
+    const otpStorage = entryData[hash];
+    let otpAuthLine = 'otpauth://';
+    if (otpStorage.type === 'totp' || otpStorage.type === 'hex') {
+      otpAuthLine += 'totp/';
+    } else if (otpStorage.type === 'hotp' || otpStorage.type === 'hhex') {
+      otpAuthLine += 'hotp/';
+    } else {
+      continue;
```
mymindstorm, line 85: This excludes steam and battle.net tokens and will cause data loss.

view/popup.html
```diff
@@ -162,6 +162,7 @@
                 <a class="button" v-on:click="showEdgeBugWarning()" v-if="isEdge()">{{ i18n.download_backup }}</a>
                 <a download="authenticator.json" v-bind:href="exportFile" class="button" target="_blank" v-else>{{ i18n.download_backup }}</a>
                 <a download="authenticator.json" v-bind:href="exportEncryptedFile" v-if="encryption.getEncryptionStatus()" class="button" target="_blank">{{ i18n.download_enc_backup }}</a>
+                <a download="authenticator.txt" v-bind:href="exportOneLineOtpAuthFile" class="button" target="_blank">{{ i18n.download_otp_backup }}</a>
```
mymindstorm: Why are we making a new button for this? It's confusing for anyone who hasn't read the discussion on the issue. Shouldn't this replace our current unencrypted backup format?

src/ui/entry.ts
```diff
@@ -195,6 +228,98 @@ async function getCachedPassphrase() {
       });
 }
 
+function getExportDataFromOTPAuthPerLine(importCode: string) {
+  const lines = importCode.split('\n');
```
mymindstorm, line 237: Will this work with other line ending formats?

src/ui/entry.ts
```diff
@@ -326,9 +453,16 @@ async function entry(_ui: UI) {
             return decryptedbackupData;
           },
       importBackupCode: async () => {
+        let exportData: {[hash: string]: OTPStorage} = {};
+        try {
+          exportData = JSON.parse(_ui.instance.importCode);
+
+        } catch (error) {
+          // Maybe one-otpauth-per line text
+          exportData = getExportDataFromOTPAuthPerLine(_ui.instance.importCode);
```
mymindstorm: This only covers the import code tab, not the file import method.

src/ui/entry.ts
```diff
@@ -195,6 +228,98 @@ async function getCachedPassphrase() {
       });
 }
 
+function getExportDataFromOTPAuthPerLine(importCode: string) {
```
mymindstorm: Importing codes with certian characters in the account name field don't import the name successfully.
```
otpauth://totp/%20:%3A%3A%3A!!!?secret=aaaabbbbcccc&issuer=%20&period=30
```

mymindstorm (dismissed):

view/popup.html
mymindstorm: Additionally, we should probably make a learn more link and put all the projects that use this format in a wiki page. E.g. "You can import backups from some other applications. [Learn more](https://github.com/Authenticator-Extension/Authenticator/wiki/Standard-OTP-Backup-Format)" and that page can have supported apps and implementation details.

src/ui/entry.ts
Sneezry: steam is not standard otpauth, and cannot be exported as otpauth URL. battle.net is an 8-digit standard otpauth, however, we do not support digit parameter currently.

view/popup.html
Sneezry: I agree the new button may confuse the user. However, as I mentioned above, I haven't found a good solution to contain steam and battle.net entries in otpauth url.

src/ui/entry.ts
Sneezry: I believe this is a broken data. The label of otpauth URL should be fully encoded, so `!` should never be there.

src/ui/entry.ts
Sneezry: It should just work with both `\n` and `\r\n`, because we make a trim with every line.

src/ui/entry.ts
Sneezry: fixed.

src/ui/entry.ts
Sneezry: fixed.

src/ui/entry.ts
mymindstorm: Must be a problem with encoding. I was able to make this without manually editing the backup file
```
otpauth://totp/?secret=QP2RFPQGO4TKEQ64K5KDYJ7KU2NNEHVO
otpauth://totp/✓ à la mode:example.com?secret=PKX5BHAPN63RQE726HWXW2VPA72AMPOG&issuer=✓ à la mode
otpauth://totp/!!!"""'''&&==++##::?secret=AFKVXHTAZZQKCHI3XSZPX5NKQRFXL3AD&issuer=!!!"""'''&&==++##:
```
image

%3A and : are not allowed in label section.

view/popup.html
mymindstorm: I think we could either
- Use our own convention E.g.: `otpauth://steam/`
- Just force people who use those steam and blizz secrets to use encrypted backups. All we need to do is make a warning popup if those types exist then. I doubt too many people use steam and blizz since you need a rooted phone to get the secret. 
- Secretly use the older format for people with those secrets and document on the learn more wiki page

Blizz secrets are exempt from above since we can just append the digits parameter if it is standard

src/ui/entry.ts
mymindstorm: Added to backlog at #287

src/ui/entry.ts
mymindstorm: Fails without colon
```
otpauth://totp/?secret=QP2RFPQGO4TKEQ64K5KDYJ7KU2NNEHVA
```

view/popup.html
mymindstorm: Wiki page is ready

src/ui/entry.ts
Sneezry: How about just remove all colons when export as otpauth urls?

mymindstorm: That works. It seems to have stopped URI encoding issuer & account fields for some reason too.

mymindstorm: Fixed

src/ui/entry.ts
mymindstorm: Fixed

view/popup.html
mymindstorm: fixed

nmarley:
@mymindstorm I really don't like this new format as it doesn't work with my established workflow. Is there any way to opt to use the old JSON format export?

mymindstorm:
@nmarley it should use both just fine. It tries to detect which format you use.

nmarley:
Thanks. I really just mean I _prefer_ to manage my backups using the JSON format. But I've found a way to just keep adding to the JSON format one for "backups" / imports.

Cheers,

PR #323, opened 2019-04-19
Webpack

mymindstorm:
- All triple slash directives removed
- Updated vue
- QR library is now slightly less buggy
- Webpack

Future goals:
- https://vuejs.org/v2/guide/typescript.html
- https://vuejs.org/v2/guide/unit-testing.html
- https://vuejs.org/v2/guide/transitions.html
- More components

Problems:
- I don't know how to make the spritesheet work in webpack, so it's manual for now.
- Need to update about page


Sneezry:
Awesome!

Sneezry (approved):

PR #354, opened 2019-07-09
Prettier

mymindstorm:
Change code style tool to prettier so that *.vue files are formatted properly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,8 @@ stages:
 jobs:
   include:
   - stage: test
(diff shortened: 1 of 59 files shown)


Sneezry:
As we removed gts from our development tool chain, do we need update readme as well?

mymindstorm:
Sure. I want to move back to GTS when they finish migrating to eslint (since that works with vue files).

Sneezry (approved):

Issue #306, opened 2019-03-23
manual entry ui improvements


mymindstorm:
<Do not use this to ask about lost codes or accounts>
<More info here: https://github.com/Authenticator-Extension/Authenticator/wiki/Lost-Codes>
<You can also open an issue by Tweeting with #AuthenticatorFeedback>

- Text boxes should be for issuer and secret, not account name and secret. It's far more common to have a different issuer than a different username. I don't really like how Google Authenticator does it that way.
- Account name should be replaced with username, since that terminology is more common (in English at least).
- An advanced settings accordion should be added so we can hide account type and period settings under it

I'm pretty sure a user complained about the first two points at least once.

firekod:
Better make 3 fields:
issuer
secret
account name

PR #357, opened 2019-07-11
manual entry improvements (#306)

mymindstorm:


Sneezry:
zh_CN/message.json should be updated as well. Maybe we should move zh_CN back to crowdin

mymindstorm:
zh-cn is on crowdin now

src/components/Popup/AddAccountPage.vue
```diff
@@ -56,22 +73,34 @@ export default Vue.extend({
       } else {
         type = this.newAccount.type;
       }
-      const entry = new OTPEntry({ 
+
+      if (type === OTPType.hhex || type === OTPType.hotp) {
+        this.newAccount.period = NaN;
```
Sneezry: `NaN` may not be a good default value here. If `period` is not available for specific types, `null` or `undefined` is a better option.

Sneezry (approved):

src/components/Popup/AddAccountPage.vue
mymindstorm: `this.newAccount.period` is typed as a number and it wouldn't let me set it as null. NaN evaluates to false so it will just treat it as if period was set to null. Is there a good reason not to use NaN?

Sneezry: `NaN`'s meaning is Not a Number. When you regard something not a number as a number (such as using `Number` function), its value will be set to `NaN`. Here, `hhex` or `hotp` doesn't has `period` property, but not has `period` of which value is not a number. `null` is all zero in bits in memory, which means nothing, or doesn't exist. We can change type of `this.newAccount.period` to `number | null`.

PR #359, opened 2019-07-13
Autolock

mymindstorm:


Sneezry:
Removing `"persistent": false` doesn't make sense.

Cookies itself has expires feature, and we can use this feature to handle auto lock.

mymindstorm:
Couldn't get it to remove the cookie from the background script, lemme revert that and try again.

mymindstorm:
Yeah, since we set the cookie in the window, it's only accessible from the window. I'm going to move all of the logic relating to the cookie to the background script.

mymindstorm:
Basically:


Problem 1: Can't access cookie from background script


Solution: Move cookie to background script


Problem 2: Timeout in background doesn't work (likely due to content script getting unloaded)


Solution: Make background script persistent

After moving everything to background and then making it un-persistent, it becomes redundant to store the passphrase in a cookie so I removed it.

Sneezry:
I think we needn't move cookie to background. When the user change auto lock settings in popup page, script runs in popup page can change the cookie. And when it's time cookie expires, browser should remove the cookie, and we do not need to do the removal in our code.

mymindstorm:
How about this?

erickufrin-okta:
@mymindstorm this autolock feature is very interesting to us! we have this need... question, is there a way which we can enforce a password and enforce auto-lock? Does this chrome extension offer the a way to control these settings via chrome policy? chrome://policy I can see Authenticator there, but no managed/manageable settings...
Screen Shot 2019-08-15 at 2 20 35 PM

mymindstorm:
@erickufrin You need to check the box that says "Show policies with no value set". I have some documentation [here](https://authenticator.cc/docs/en/policy-config), look at that schema file for more info.

Next update has a lot of changes, make an issue for it and I'll try to fit some policies in for you.

Sneezry (approved):

PR #361, opened 2019-07-13
Keyboard nav

### LINKED ISSUE #269: Please add a keyboard shortcut!
Brilliant little thing you made, thanks very much.

There are many instances (at least in my case) where the Chrome window does not have url/extension toolbar (e.g. popups). It would be great to have a keyboard shortcut to bring up Authenticator and directly into focus so we can use the keyboard to select and copy the TOTP, then right back to the webpage to paste it.

Generally speaking, it would be great to be able to do everything from the keyboard quickly. This helps in situations where things need to be done quickly, e.g. stock or currency exchanges which ask for TOTP every 1h when placing orders, and some of those times are very critical (you don't want to waste seconds).

Many thanks in advance!

mymindstorm:
Todo:
- [ ] Add setting page nav
- [ ] Label exit and back nav buttons
- [ ] Test with screen reader
- [ ] Notify screen reader on copy
- [ ] For general users focus entry on up / down keys

closes #269

### DIFF
--- a/sass/popup.scss
+++ b/sass/popup.scss
@@ -548,6 +548,7 @@ svg {
 // Info
 #info {
   position: absolute;
+  display: none;
   height: 460px;
   width: 300px;
   padding: 10px;
@@ -558,7 +559,7 @@ svg {
     box-shadow: 1px 1px 3px themed("grey-1");
   }
   left: 10px;
-  top: -1000px;
+  top: 10px;
   z-index: 100;
 
   #infoContent {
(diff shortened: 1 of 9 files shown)


sass/popup.scss
```diff
@@ -548,6 +548,7 @@ svg {
 // Info
```
saipul-anwar21a: https://web.facebook.com/checkpoint/?next=https%3A%2F%2Fweb.facebook.com%2Fr.php%3Flocale%3Did_ID

Issue #369, opened 2019-07-30
Replace MD5 with SHA256


PatOnTheBack:
**Describe the bug:**
Authenticator used MD5, but MD5 is vulnerable to collision attacks (https://en.wikipedia.org/wiki/MD5#Collision_vulnerabilities).
Replacing MD5 with SHA256 will mitigate this risk.

**Platform:** All
 - Browser: All
 - Browser Version: All

mymindstorm:
closed by #395

PR #370, opened 2019-07-30
Replace MD5 with SHA256

PatOnTheBack:
This addresses issue #369.

### DIFF
--- a/src/background.ts
+++ b/src/background.ts
@@ -167,7 +167,7 @@ async function getTotp(text: string) {
         chrome.tabs.sendMessage(id, { action: "secretqr", secret });
       } else {
         const encryption = new Encryption(cachedPassphrase);
-        const hash = CryptoJS.MD5(secret).toString();
+        const hash = CryptoJS.SHA256(secret).toString();
         if (
           !/^[2-7a-z]+=*$/i.test(secret) &&
           /^[0-9a-f]+$/i.test(secret) &&
--- a/src/components/Popup/AddAccountPage.vue
+++ b/src/components/Popup/AddAccountPage.vue
@@ -62,7 +62,7 @@ export default Vue.extend({
         issuer: "",
         account: this.newAccount.account,
         encrypted: false,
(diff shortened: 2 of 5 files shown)


mymindstorm:
Thanks for the PR! I'm 99% sure this will break all existing backups and encryption. This needs to migrate old data.

This has been on my to-do list for a while. Instead of just moving to SHA256, it might be a better idea to move to a full blown password hashing algorithm via libsodium or some other crypto library that's being actively maintained.  

I'm on vacation right now, I'll need a while to get around to this (and finishing up 5.6).

Sneezry:
We use MD5 as an identity of an entry, but not for data protection. Will this issue affect us? What's your idea @mymindstorm

PatOnTheBack:
I suppose it is more of a theoretical threat. Someone who is manipulating the code in the browser extension could create a fake entry that has the same MD5 hash as the real entry.
The infosec industry is gradually moving away from MD5 because of the vulnerabilities discovered in it.
SHA256 will run more slowly than MD5, so if speed is important, MD5 might still be a better choice.

PatOnTheBack:
@mymindstorm I agree that using a password hashing algorithm is best.
libsodium only supports Scrypt and Argon2.
This [NPM package for bcrypt](https://www.npmjs.com/package/bcrypt) seems to be currently maintained, so bcrypt is another option.

mymindstorm:
> Someone who is manipulating the code in the browser extension could create a fake entry that has the same MD5 hash as the real entry.

I don't really see the point of moving solely due to the collision vulns. The most damage exploiting that (in our context) could do would just replace an entry w/ bad data. My understanding of the benefit of a password hashing algorithm is that they are intentionally slow as to make brute forcing infeasible.

<details>
<summary>For me</summary>

https://github.com/antelle/argon2-browser

https://www.npmjs.com/package/@types/argon2-browser
</details>

Sneezry:
I'm closing this PR because we have no plan to change the hash currently to avoid breaking change. Please feel free to raise a new thread if you have any other idea. Thanks for your contribution.

Issue #366, opened 2019-07-18
Allow multiple entires using same secret


akiller:
**Describe the feature you want:**
It would be great if you were able to add multiple entries using the same secret key. E.g., consider the following examples:

```
otpauth://totp/example.com:akiller?secret=JBSWY3DPEHPK3PXP&issuer=example.com&algorithm=SHA1&digits=6&period=30
otpauth://totp/localhost:akiller?secret=JBSWY3DPEHPK3PXP&issuer=localhost&algorithm=SHA1&digits=6&period=30

```

Currently, if you add the first one then add the second the first one is deleted (or updated), and only the second entry remains.

An alternative (and possibly preferred) approach would be to work how LastPass does, and allow you to edit an entry and supply a list of alternative domains that it will match and autofill on.

**Why do you want this feature in Authenticator?**
When using the the autofill feature it would be nice to have it work across multiple domains. Especially when using a multitude of different staging / testing / production domains that use the same secret key for the user account.

mymindstorm:
Append `::localhost` to the issuer to manually override autofill. It should match for the live and local version of the site. We store based on hash, if you try to create another entry with the same hash as one already in storage it will just supersede the one in storage.

akiller:
Thanks for the suggestion I'll give that a try. 

You wouldn't believe how confused some people look when you ask them to type in six numbers so I'm just trying to make it as seamless as possible. This browser extension has definitely helped a lot!

PR #377, opened 2019-08-19
Argon2

mymindstorm:
Fixes #370 and #366


mymindstorm:
I think I'm going to change how this works. Instead of directly encrypting each entry, it would be better to generate a random key, encrypt the entries with that key, encrypt the key with the user's password, and only check the hash of the key.

PR #385, opened 2019-08-30
More policies

### LINKED ISSUE #375: Policy: Autolock + Require Password
Allow for centralized policy control of certain settings + enforcement of password.

- enforce autolock after X-num min (for flexibility it should allow for X-number of minutes vs pre-defined increments)
- enforce autolock on startup (upon launch of a new chrome tab, the OTP's should be locked)
- enforce password be set (dont allow enrolling OTP tokens without a password set first)

Without these features it is easy for someone to access OTP codes on an unlocked workstation.

For things which may only require an OTP code, by having a password to unlock the Authenticator extension gives us an additional "factor" of "something known" as well as "something I have" being the enrolled OTP extension itself.

While having optional features to autolock and passwords is great, we need to have a way to make them mandatory by policy and push via normal management channels.

The 3 ways we apply policy to Chrome extensions are:

- Windows Group Policy
- macOS JAMF
- ChromeOS cPanel

mymindstorm:
Closes #375

### DIFF
--- a/src/background.ts
+++ b/src/background.ts
@@ -25,7 +25,6 @@ chrome.runtime.onMessage.addListener((message, sender, sendResponse) => {
       message.info.windowWidth
     );
   } else if (message.action === "cachePassphrase") {
-    document.cookie = `passphrase="${message.value}${getCookieExpiry()}"`;
     cachedPassphrase = message.value;
     clearTimeout(autolockTimeout);
     setAutolock();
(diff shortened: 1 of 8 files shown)


src/background.ts
```diff
@@ -405,25 +404,37 @@ chrome.commands.onCommand.addListener(async (command: string) => {
   }
 });
 
-function setAutolock() {
+async function setAutolock() {
+  const enforcedAutolock = Number(await ManagedStorage.get("enforceAutolock"));
+
+  if (enforcedAutolock) {
+    if (enforcedAutolock > 0) {
+      document.cookie = `passphrase="${getCachedPassphrase()}${getCookieExpiry(
```
Sneezry, line 412: Haven't we discarded cookies in the recent design?

mymindstorm: That's only part of #395 currently. There are some changes to how we do cookies that the PR doesn't have yet, so I'm going to go through and remove them again after merge.

Sneezry (approved):

PR #398, opened 2019-10-27
QR scan encryption fix

mymindstorm:
Adding QR code when encryption is empty will add unencrypted data to storage. This can happen if scanning a QR code after autolock triggers.


Sneezry:
Will autolock be triggered while popup page is open?

mymindstorm:
Yes, but it doesn't matter since popup makes a copy of encryption on load. The same applies to import.

Sneezry (approved):

PR #413, opened 2019-12-24
fix #412

### LINKED ISSUE #412: Fix for secure storage in Chrome
This is followup to #384, sorry for taking too long... i hope/dont know if it is still relevant with latest browser updates. I was trying this about an half year ago, but my version of extension is 5.3.2

In Chrome there is (and should be still) problem with storing anything in local storage,
there is "transaction log" 000003.log with everything what was written and even overwritten/deleted! It is described in detail here: https://blog.bitwarden.com/chrome-extension-version-1-24-security-fix-1ce700aeccf6

So issue is when enabling encryption in your extension if there were some unencrypted entries before, it will stay in the file! Chrome may cleanup this file some time in future or fix this or who knows... Workaround fix is when saving delete everything that was stored before, all these values are somehow in memory already so nothing is lost only old is erased... I dont know if this is feature of your code or Chrome storage api, but please refactor so it is clear what methods do....

I somehow hacked it but this JS is hacky on its own, i dont even know what it is exactly doing as of now. Try for yourself and refactor! PS I am not sure if my approach does not erase the whole user storage, but it should be limited to this extension as of storageAPI.
There is much duplicity and crossing of concerns in what methods do like set/remove.... add/update/set there is two classes in one file and methods with 100 lines are not good.
While dissecting EntryStorage and BrowserStorage to distinct classess/files try to introduce for example polymorphic separation (2 classes with common interface) for "local" storage and "sync" storage and get rid of those if statements on every other line...

https://refactoring.guru/replace-conditional-with-polymorphism
Also relevant is SingleResponsiblity, so for example BrowserStorage getStorageLocation should not check wheter browser is Edge or whatever... that belongs to some other (extracted) function

Here:[storage.js](https://github.com/Authenticator-Extension/Authenticator/files/3998624/storage.txt)

mymindstorm:


src/models/storage.ts
```diff
@@ -114,6 +114,25 @@ export class BrowserStorage {
     }
     return;
   }
+
+  // Use for Chrome only.
+  // https://github.com/Authenticator-Extension/Authenticator/issues/412
+  static async clearLogs() {
```
Sneezry, line 120: async won't work as expected because we use callback in the function. I found this project, and it seems a good option for us: https://github.com/Lusito/webextension-polyfill-ts

mymindstorm: I'm only using the async to make `await this.getStorageLocation` work properly. Moving to that polyfill should be another PR entirely since it would change so much.

mymindstorm: + wouldn't this not work anywhere we have chrome specific calls?

Sneezry: It seems work with Chrome: https://github.com/mozilla/webextension-polyfill#supported-browsers

src/store/Accounts.ts
```diff
@@ -335,8 +335,13 @@ export class Accounts implements IModule {
             state.state.encryption.updateEncryptionPassword(
               wordArray.toString()
             );
+
             await state.dispatch("updateEntries");
 
+            if (navigator.userAgent.indexOf("Chrome") !== -1) {
```
Sneezry, line 342: How about add a bug link here to remind why we should clear logs for Chrome

Sneezry (approved):

src/models/storage.ts
mymindstorm: fixed async issue. I agree that we should move to webextension-polyfill.

PR #419, opened 2020-01-15
Fix #414

### LINKED ISSUE #414: Can't export after setting a password
Added a password in Security. Now When I try to click Import/Export I have a popup 

>You cannot add a new account or export data until all accounts are decrypted. Please enter the correct password before continuing

**Platform:** Windows 10
 - Browser: Chrome
 - Browser Version: Version 79.0.3945.88 (Official Build) (64-bit)

mymindstorm:


src/models/encryption.ts
```diff
@@ -26,20 +26,6 @@ export class Encryption implements IEncryption {
         return null;
       }
 
-      if (decryptedSecret.length < 8) {
```
Sneezry: Should we remove the check here? Though we have already checked that when users add new accounts, it would be good for better exception handling.

mymindstorm: I didn't want to break any codes that don't follow the RFC since the check should be for 16 chars. If we do add a stricter check it shouldn't be in this function anyway. It should make the code show invalid [around here](https://github.com/Authenticator-Extension/Authenticator/blob/fix-encrypted/src/models/otp.ts#L134).

mymindstorm: I didn't want to break any codes that don't follow the RFC since the check should be for 16 chars. It is a good idea to keep this and probably add another that would make the code show "Invalid" if it isn't at least 16 chars long.

PR #429, opened 2020-02-02
improve password change and migrate logic

### LINKED ISSUE #428: Cannot decrypt some accounts after v6.0 update
Hi,
after the latest update v6.0 when I open the extension and I unlock it with password, not all my accounts are visible. Some of them remain encrypted.
Moreover, even if I setup a lock after some minutes, authenticator locks each time I close it.
I'm using the latest firefox v72.0.2 on GNU/linxu 64 bit.

mymindstorm:
fix #428

### DIFF
--- a/src/store/Accounts.ts
+++ b/src/store/Accounts.ts
@@ -351,7 +351,7 @@ export class Accounts implements IModule {
               key: { enc: encKey, hash: encKeyHash }
             });
             await EntryStorage.set(state.state.entries);
-            if (removeHashes) {
+            if (removeHashes.length) {
               await new Promise(resolve => {
                 BrowserStorage.remove(removeHashes, () => {
                   resolve();
--- a/src/store/Accounts.ts
+++ b/src/store/Accounts.ts
@@ -336,7 +336,11 @@ export class Accounts implements IModule {
                 new Encryption(wordArray.toString())
               );
               // if not uuidv4 regen
-              if (/[0-9A-F]{8}-[0-9A-F]{4}-4[0-9A-F]{3}-[89AB][0-9A-F]{3}-[0-9A-F]{12}/i.test(entry.hash)) {
+              if (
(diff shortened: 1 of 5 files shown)


src/store/Accounts.ts
```diff
@@ -320,16 +329,34 @@ export class Accounts implements IModule {
...
+                await entry.genUUID();
+              }
+            }
+
+            // store key
+            await BrowserStorage.set({
+              key: { enc: encKey, hash: encKeyHash }
+            });
+            await EntryStorage.set(state.state.entries);
+            if (removeHashes) {
```
Sneezry: This could be always true.

```js
console.log(!![]); // true
```

mymindstorm:
```suggestion
            if (removeHashes.length) {
```

src/store/Accounts.ts
```diff
@@ -320,16 +329,34 @@ export class Accounts implements IModule {
...
             // change entry encryption and regen hash
+            let removeHashes: string[] = [];
             for (const entry of state.state.entries) {
               await entry.changeEncryption(
                 new Encryption(wordArray.toString())
               );
-              await entry.genUUID();
+              // if not uuidv4 regen
+              if (
+                !entry.hash.match(
```
Sneezry: Use `test` here makes more sense

src/store/Accounts.ts
```diff
@@ -320,16 +329,34 @@ export class Accounts implements IModule {
...
+                  /[0-9A-F]{8}-[0-9A-F]{4}-4[0-9A-F]{3}-[89AB][0-9A-F]{3}-[0-9A-F]{12}/i
+                )
+              ) {
+                removeHashes.push(entry.hash);
+                await entry.genUUID();
```
Sneezry: `genUUID` is a sync method

Sneezry (approved):

PR #434, opened 2020-02-09
OneDrive Support

### LINKED ISSUE #346: OneDrive for backup sync
<You can also suggest feedback by Tweeting with #AuthenticatorFeedback>

We use Onedrive and it would be great if you could add that to your list of cloud backup services in the app.

I want it so that we can utilize the cloud sync backup feature as we do not use dropbox or google.

Thank you.

mymindstorm:
Closes #346 

To access credentials go to Azure => switch to mymindstorm.onmicrosoft.com tenant => App Registrations. You will need to create a testing client secret. Redirect URI will not match if key is not set correctly in manifest.


mymindstorm:
The domain verification system won't verify authenticator.cc because github pages adds `charset=utf-8` to the end of the content-type header. Yay.

edit: Fixed it. You just need to go to AAD's custom domain settings and verify it using their completely different system.

src/models/backup.ts
```diff
@@ -460,3 +459,204 @@ export class Drive implements BackupProvider {
...
+    }
+    const exportData = await EntryStorage.backupGetExport(
+      encryption,
+      localStorage.oneDriveEncrypted === "true"
+    );
+    const backup = JSON.stringify(exportData, null, 2);
+
+    const token = await this.getToken();
+    if (!token) {
+      return new Promise((resolve: (value: boolean) => void) => {
```
Sneezry: Here should be safe to just return false instead of a promise

src/models/backup.ts
```diff
@@ -460,3 +459,204 @@ export class Drive implements BackupProvider {
...
+        resolve(false);
+      });
+    }
+
+    return new Promise(
+      (resolve: (value: boolean) => void, reject: (reason: Error) => void) => {
+        if (!token) {
+          resolve(false);
```
Sneezry: Suggestion: I prefer to always add return before resolve and reject, because resolve and reject pass the result of the async function, and the function should end there. Always add return where a promise resolved or rejected can avoid possible bugs

```
return resolve(false);
```

src/models/backup.ts
```diff
@@ -460,3 +459,204 @@ export class Drive implements BackupProvider {
...
+            "PUT",
+            `https://graph.microsoft.com/v1.0/me/drive/special/approot:/${now}.json:/content`
+          );
+          xhr.setRequestHeader("Authorization", "Bearer " + token);
+          xhr.setRequestHeader("Content-type", "application/octet-stream");
+          xhr.onreadystatechange = () => {
+            if (xhr.readyState === 4) {
+              if (xhr.status === 401) {
+                localStorage.removeItem("oneDriveToken");
+                resolve(false);
```
Sneezry: The code would contine to run event we have resolve the promise here

src/models/backup.ts
```diff
@@ -460,3 +459,204 @@ export class Drive implements BackupProvider {
...
+          return reject(error);
+        }
+      }
+    );
+  }
+
+  async getUser() {
+    const token = await this.getToken();
+    if (!token) {
+      return new Promise((resolve: (value: string) => void) => {
```
Sneezry: Should we resolve the promise here? It seems reject makes more sense here. Also, here is safe to return the string directly.

mymindstorm:
I went through all of the resolves and rejects and fixed the returns there too.

Sneezry (approved):

Issue #461, opened 2020-04-22
Don't show QR code button on entry when disableExport is enabled


codycalder:
Is it possible to access a version of this Chrome extension that does not allow users to click to view the QR code? I really like this extension and it works great for my requirements however I do not want my users to be able to see the QR code as this allows them to load on another authenticator device. Thanks in advance! 
![Screen Shot 2020-04-22 at 5 46 27 pm](https://user-images.githubusercontent.com/57397772/79957159-86e78e00-84c4-11ea-864c-79e9090d4f66.png)

codycalder:
Hi @Sneezry - That would be perfect, yes!

PR #465, opened 2020-04-23
disable qr code when set disableExport to true

Sneezry:
https://github.com/Authenticator-Extension/Authenticator/issues/461


mymindstorm (changes requested):
Instead of using `store/Entry.ts`, use the variable `exportDisabled` in `store/Menu.ts`

Sneezry:
Done.

src/store/Menu.ts
```diff
@@ -13,6 +13,7 @@ export class Menu implements Module {
         exportDisabled: await ManagedStorage.get("disableExport"),
         enforcePassword: await ManagedStorage.get("enforcePassword"),
         enforceAutolock: await ManagedStorage.get("enforceAutolock"),
+        showQrDisabled: await ManagedStorage.get("disableExport"),
```
mymindstorm: Why did you add a line here? The exact same thing happens on line 13

Sneezry: Ooooops

mymindstorm (approved):

PR #466, opened 2020-04-23
fix #464

### LINKED ISSUE #464: Can't type my password if it contains /
**Describe the bug:**

Recently I noticed that I can no longer type in my password. As soon as type in `/` the focus is taking away from the password prompt into the search bar.

**Platform:**
 - Browser: Chrome on Linux and Chrome on Mac.
 - Browser Version: `81.0.4044.92` on Linux and `81.0.4044.113` on Mac.

mymindstorm:


Sneezry:
Use event.preventDefault in input box key event to stop bubbling makes more sense. Condition in this PR is too complex

mymindstorm:
How's this instead? We can use event.preventDefault, but I'd rather not since we would have to use it on every text input we add.

Sneezry:
We can also check keyup event target to consider show search box or not

mymindstorm:
If you copy a code it focuses an input and breaks search with 539c615

https://github.com/Authenticator-Extension/Authenticator/blob/0f8294389a90089b31f4fa77cae027081f580397/src/components/Popup/EntryComponent.vue#L207

Sneezry (approved):

PR #488, opened 2020-05-22
update about page

mymindstorm:


scripts/build.sh
```diff
@@ -59,6 +59,12 @@ fi
 ./node_modules/sass/sass.js sass:css
 cp ./sass/DroidSansMono.woff2 ./css/
 
+echo "Generating licenses file..."
+./node_modules/.bin/npm-license-generator \
+    --out-path ./view/licenses.html \
+    --template ./scripts/licenses-template.html \
+    --error-missing=true
```
Sneezry, line 66: <del>`--registry` should be a requied parameter right?</del>

Sneezry: also, I get `URL` is not defined error from `Authenticator/node_modules/npm-license-generator/build/index.js:100:21`

Sneezry (changes requested):

scripts/build.sh
mymindstorm: registry is optional and defaults to registry.npmjs.com. What version of node are you using?

Sneezry: It's Node 8. I'm upgrading to 12 to test again.

mymindstorm: URL was considered experimental in node 8, I need to add an import for it.

Sneezry: I have verified on Node 12, no errors in build.

Sneezry (approved):

scripts/build.sh
mymindstorm: node 8 is supported with 1.0.1

PR #494, opened 2020-05-29
Add security policy

mymindstorm:
We should have a way to report serious security issues. I can add you to the mailing list if you want.


Sneezry:
Is vulnerability report neccessary for open source project? I mean vulnerability is technology level, only customers with tech skill may find vulnerability. And if the customer can find it, then should also be able to make a PR to fix it.

PR #496, opened 2020-06-08
Add security key support

Sneezry:


_locales/en/messages.json
```diff
@@ -353,5 +353,20 @@
     },
     "algorithm": {
         "message": "Algorithm"
+    },
+    "operationFailed": {
+        "message": "Operation failed."
```
mymindstorm, line 358: Crowdin is going to flood my email again :(

use `updateFailure`

_locales/en/messages.json
```diff
@@ -353,5 +353,20 @@
...
+    },
+    "noPassphraseWarning": {
+        "message": "You must set a passphrase before enable security key."
```
mymindstorm, line 361:
```suggestion
        "message": "You must set a password before adding a security key."
```

_locales/en/messages.json
```diff
@@ -353,5 +353,20 @@
...
+    },
+    "securityKeyEnabled": {
+        "message": "Security key enabled."
```
mymindstorm, line 364:
```suggestion
        "message": "Security key added."
```

src/components/Popup/EnterPasswordPage.vue
```diff
@@ -29,12 +40,31 @@ export default Vue.extend({
   computed: {
     wrongPassword() {
       return this.$store.state.accounts.wrongPassword;
+    },
+    securityKeyEnabled() {
+      return !!localStorage.securityTokenEncryptedKey;
```
mymindstorm, line 45: `!!`?

_locales/en/messages.json
```diff
@@ -353,5 +353,20 @@
...
+    },
+    "alreadyEnabledSecurityKeyWarning": {
+        "message": "You have already enabled security key, do you want to continue?"
```
mymindstorm, line 367:
```suggestion
        "message": "Your old security key will not work, are you sure you want to continue?"
```

_locales/en/messages.json
```diff
@@ -353,5 +353,20 @@
...
+    },
+    "updatePassphraseDisableSecurityKeyWarning": {
+        "message": "Update passphrase will reset security key, do you want to continue?"
```
mymindstorm, line 370:
```suggestion
        "message": This action will also remove your security key, do you want to continue?"
```

src/store/Accounts.ts
```diff
@@ -395,6 +395,9 @@ export class Accounts implements Module {
 
           // remove cached passphrase in old version
           localStorage.removeItem("encodedPhrase");
+
+          // remove security token encrypted key in old version
+          localStorage.removeItem("securityTokenEncryptedKey");
```
mymindstorm, line 400: There is no older version that uses this key?

svg/windows-hello.svg
```diff
@@ -0,0 +1 @@
...-.159.373 0 .706.086 1 .258.293.172.546.41.76.714 1.253 1.798 2.723 3.166 4.41 4.104 1.686.939 3.656 1.408 5.91 1.408 2.133 0 4.043-.492 5.73-1.477 1.686-.985 3.123-2.33 4.31-4.035.2-.291.436-.525.71-.704.273-.178.596-.268.97-.268.6 0 1.106.185 1.52.555.413.371.62.866.62 1.487 0 .198-.027.38-.08.545a2.382 2.382 0 01-.24.506 8.897 8.897 0 01-1.15 1.626c-.447.503-.91.985-1.39 1.447a15.888 15.888 0 01-5.2 3.281A16.654 16.654 0 0122.38 34zM18.3 14.252c0 .436-.087.853-.26 1.249a3.423 3.423 0 01-.7 1.041 3.292 3.292 0 01-1.03.704 3.138 3.138 0 01-1.27.258c-.454 0-.877-.086-1.27-.258a3.308 3.308 0 01-1.73-1.715 3.052 3.052 0 01-.26-1.259c0-.449.083-.872.25-1.269a3.24 3.24 0 011.72-1.745A3.131 3.131 0 0115.02 11c.453 0 .876.086 1.27.258a3.39 3.39 0 011.75 1.735c.173.39.26.81.26 1.259zm8.4 0c0-.449.086-.872.26-1.269.173-.397.406-.74.7-1.031.293-.291.64-.522 1.04-.694S29.526 11 29.98 11c.453 0 .876.086 1.27.258a3.275 3.275 0 011.72 1.735c.166.39.25.81.25 1.259 0 .45-.087.869-.26 1.259-.174.39-.407.734-.7 1.031a3.292 3.292 0 01-1.03.704 3.138 3.138 0 01-1.27.258c-.454 0-.877-.086-1.27-.258a3.252 3.252 0 01-1.03-.704 3.426 3.426 0 01-.7-1.031 3.052 3.052 0 01-.26-1.259z" fill="#0078d4"/></svg>
```
mymindstorm, line 1: Do we need to add lines in the license file for these?

src/store/Accounts.ts
```diff
@@ -483,6 +486,138 @@ export class Accounts implements Module {
               });
             });
           }
+        },
+        enableSecurityKey: async (state: ActionContext<AccountsState, {}>) => {
+          const createCredentialOptions: CredentialCreationOptions = {
+            publicKey: {
+              rp: {
+                id: chrome.runtime.id,
```
mymindstorm, line 494: This is different between browsers. Will this break backups between Chrome <=> Firefox if enabled?

src/store/Accounts.ts
```diff
@@ -483,6 +486,138 @@ export class Accounts implements Module {
               });
             });
           }
+        },
+        enableSecurityKey: async (state: ActionContext<AccountsState, {}>) => {
+          const createCredentialOptions: CredentialCreationOptions = {
+            publicKey: {
+              rp: {
```
mymindstorm, line 493: We can add an icon! https://developer.mozilla.org/en-US/docs/Web/API/PublicKeyCredentialCreationOptions/rp

src/store/Accounts.ts
```diff
@@ -483,6 +486,138 @@ export class Accounts implements Module {
...
+                id: chrome.runtime.id,
+                name: "Authenticator"
+              },
+              user: {
+                name: "Authenticator",
```
mymindstorm, line 498: Is it possible to remove the name / displayname since we don't actually know any of this info?

src/store/Accounts.ts
```diff
@@ -483,6 +486,138 @@ export class Accounts implements Module {
...
+                displayName: "Authenticator",
+                id: new Uint8Array(16)
```
mymindstorm, line 500: Is it really ok for us to just have blank ids?

src/store/Accounts.ts
```diff
@@ -483,6 +486,138 @@ export class Accounts implements Module {
...
+              },
+              challenge: new Uint8Array(16),
```
mymindstorm, line 502: this should have some random data. we should verify that the key is working and response is signed properly

src/store/Accounts.ts
```diff
@@ -483,6 +486,138 @@ export class Accounts implements Module {
...
+              pubKeyCredParams: [
+                { type: "public-key", alg: -7 },
+                { type: "public-key", alg: -35 },
+                { type: "public-key", alg: -36 },
+                { type: "public-key", alg: -257 },
```
mymindstorm, line 507: This is deprecated

https://www.iana.org/assignments/cose/cose.xhtml#algorithms

src/store/Accounts.ts
```diff
@@ -483,6 +486,138 @@ export class Accounts implements Module {
...
+                { type: "public-key", alg: -258 },
```
mymindstorm, line 508: Not recommended

https://www.iana.org/assignments/cose/cose.xhtml#algorithms

src/store/Accounts.ts
```diff
@@ -483,6 +486,138 @@ export class Accounts implements Module {
...
+                { type: "public-key", alg: -259 },
```
mymindstorm, line 509: Not recommended

https://www.iana.org/assignments/cose/cose.xhtml#algorithms

mymindstorm (changes requested):
Please ignore all the other review comments and the TODOs below for now.

This just encrypts the user's password directly with the credential id. The credential id isn't secure (I.e. we can't guarantee that it's random data). We can sign and verify data, but webauthn does not provide any way to securely encrypt data (it looks as if only client-server flows were considered). It seems to be possible using native code, but that would probably be too complicated from a user's perspective. I found lots of good discussion on encryption using fido keys below. It looks like our best bet would be to propose changes to the WebAuthn API...

https://groups.google.com/a/fidoalliance.org/d/topic/fido-dev/kMe6Kb4GCpU/discussion
https://bugs.chromium.org/p/chromium/issues/detail?id=1023225#c7
https://bugs.chromium.org/p/chromium/issues/detail?id=1023225#c12
https://fidoalliance.org/specs/fido-v2.0-ps-20190130/fido-client-to-authenticator-protocol-v2.0-ps-20190130.html#sctn-hmac-secret-extension


<details>
<summary>Old review comments</summary>

- WebAuthn supports linux
- check if webauthn is supported on the browser before showing the button https://caniuse.com/#search=webauthn / MDN
- ~~Can we store the `securityTokenEncryptedKey` in `key` so that it is stored in backups? How will backups work with this in general~~

</details>

mymindstorm:
There is already a draft: https://github.com/w3c/webauthn/pull/1424

Sneezry:
> This just encrypts the user's password directly with the credential id.

Correct. Credential id is generated with the hardware, and it's value can be gotten only when the user touch the figure scanner or type the PIN. Though WebAuthn doesn't provide a way to encrypt data, but we also never expose the credential id, so I don't think there is a risk.

> It seems to be possible using native code

No, we shouldn't touch native code. It's beyond our scope. And we can't maintain native code for all platforms.

mymindstorm:
> Correct. Credential id is generated with the hardware, and it's value can be gotten only when the user touch the figure scanner or type the PIN. 

Unfortunately, this is not the case. There is nothing preventing a vendor from making the id a hash of publicly known information or a serial that is incremented every time a new credential is generated. In either case an attacker would be able to get plaintext user password with current implementation.

The below email thread is a good example of a security key vendor about to do that. It's worth noting that the Yubico dev says that it's completely up to the vendor how this is implemented.

https://groups.google.com/a/fidoalliance.org/d/topic/fido-dev/5mwvgna3RKc/discussion

Below is a link to the spec, the "key handle" is the rawId. Same as above, there is nothing explicitly preventing an easily guessable id.

https://fidoalliance.org/specs/fido-u2f-v1.1-id-20160915/fido-u2f-raw-message-formats-v1.1-id-20160915.html#registration-response-message-success

> Though WebAuthn doesn't provide a way to encrypt data, but we also never expose the credential id, so I don't think there is a risk.

As I understand it, normally you are supposed to store the credential id somewhere [and use it when calling credentials.get](https://developer.mozilla.org/en-US/docs/Web/API/PublicKeyCredentialRequestOptions/allowCredentials). Without storing the id, there is no guarantee that we will get the same credential if the user ever regenerates their key.

I recently found that some browsers already have an hmacCreateSecret extension we could use to do this in a secure way. It seems that it was made specifically for Windows Hello support. What do you think about using that?

https://groups.google.com/a/fidoalliance.org/d/topic/fido-dev/SX6AoLAoOx8/discussion

Sneezry:
> Unfortunately, this is not the case...

There isn't a standard for how to generate credential id hash, so the attacker also has no idea how to generate the credential id with RP id. Also, user id is a random number that only use once when we register the device. We don't store user id, either. The way we use webauthn here is very different from how it is designed. We don't store anything but the encrypted secret with the credential id. The only thing can be known without customer's permission is RP id.

> As I understand it, normally you are supposed to store the credential id somewhere...

The way we use webauthn here is very different from how it is designed, we don't store the credential id. The credential id is stored in the hardware security chip. Instead of using `allowCredentials`, which requires the credential id, we just list all credentials match our RP id, and ask customer to choose the correct one - even if the customer has the correct PIN, if the chosen credential is not the one to encrypt the password, decyption will fail (for example, customer use Windows Hello to encrypt data, however, chose another hardward security token to decrypt, will meet an error).

---

I understand webauthn is not designed for encryption. However, if the credential id cannot be read without customer's permission, it can be a good key to protect the data.

src/components/Popup/EnterPasswordPage.vue
Sneezry: `!!localStorage.securityTokenEncryptedKey` always has bool value

src/store/Accounts.ts
Sneezry: I thought it will be some random chars without setting a value just like C language. I will update to use a random string.

src/store/Accounts.ts
Sneezry: We don't care about the credential, we are only intereted in credential id

src/store/Accounts.ts
Sneezry: It's the only supported algorithm for Windows Hello. https://docs.microsoft.com/en-us/microsoft-edge/dev-guide/windows-integration/web-authentication

mymindstorm:
I have thought about this for the past few days, and I am still not comfortable with this. I do not think that using a standard in a non-intended way is a smart move when dealing with sensitive user data. We do not have any guarantees nor can we make any assumptions about the entropy of the key handle.

If you still want to move forward with this and keep the same encryption scheme, we should send an email to the fido-dev mailing list and get an opinion from the implementors and developers of the spec.

Sneezry:
> I do not think that using a standard in a non-intended way is a smart move when dealing with sensitive user data.

I totally agree with you.

> If you still want to move forward with this and keep the same encryption scheme...

We have no need to move forward in such a hurry. This is not an urgent or required feature. I just publish it for review and discussion. Let's make everything reliable and ready before ship it.

Sneezry:
>I know that HMAC secret description is misleading, but you can not use it for "encryption". The extension is generating HMAC key, scoped to the credential. During GetAssertion you can provide HMAC secret extensions with one or two salts, and you will be given HMAC for each of the salts. Platform can leverage that to do offline authentication by rolling salts.
>
>And yes, browser are not supporting it.
>
>Pleaser read HMAC secret extension definition properly https://fidoalliance.org/specs/fido-v2.0-id-20180227/fido-client-to-authenticator-protocol-v2.0-id-20180227.html#sctn-hmac-secret-extension.

https://groups.google.com/a/fidoalliance.org/d/msg/fido-dev/kMe6Kb4GCpU/EZjN19qhBQAJ

It seems HMAC extension is not what we want

mymindstorm:
I missed that, I was mistaken. After re-reading that thread, it is clear that we should not be using WebAuthn for encryption at all.

Sneezry:
Correct. We have to wait for new web standard for data encryption scenario. I'm closing this PR, and keep an eye on the change in W3C org.

Issue #510, opened 2020-07-19
The permission automatically given is file:///*


MHGhandour:
<Do not use this to ask about lost codes or accounts>
<More info here: https://authenticator.cc/docs/en/lost-codes>

**Describe the bug:** <Describe the issue with Authenticator> 

**Platform:** <This is required>
 - Browser: [Chrome, ]
 - Browser Version: Brave

Hello,
I noticed when i managed the extension that it has the permission to read file:///* which basically mean it can read all my local files, including my files that contains all my saved passwords.

![image](https://user-images.githubusercontent.com/55106045/87867470-3ed5eb80-c98d-11ea-801a-b63c1ec2888b.png)

Sneezry:
The `file:///*` permission is not like what you think. Let me explain what the ability this permission grants, and why we need this permission.

First of all, Chrome extension is not like native apps run on your machine - it is not able to read your local files autonomously. If the extension needs to read your local file, these three conditions must meet:

1. You enable `file:///*` host permission in extension management dashboard (which is diasbled in your screenshot)
2. You must open that local file in the browser
3. You must make the tab active, which opens the local file, and click the extension icon

We need this permission, because some users save the QR image locally, when they need import data with the local QR image file, the extension must need `file:///*` permission to access the local image.

Though the extension declares `file:///*` permission, that permission must be granted by the user in extension management dashboard manually. The permision is always disabled by default, this is as designed by Google.

MHGhandour:
Ok, Thank you for explaining that. i thought that the access did not require me opening the directory in the tab.  
Although, for me it was on by default, i disabled it. 
i don't know if it is possible to change this directory to something else set by the user, like ~/Pictures for example, Cause this seemed scary at first to me before realizing it's opensource. 

Anw, thank you for this extension, it is well needed

Sneezry:
Did you install this extension before or is this your first time to install it? Did you enable it before? If this permission is enbaled by default, it's a serious security issue to us.

If Google has changed the behaviour, we need change it to an optional permission.

Here's an [answer](https://stackoverflow.com/a/19493206) on StackOverflow to prove such a behaviour. I'm not sure if the answer has been out of date.

Sneezry:
> i thought that the access did not require me opening the directory in the tab.

I think it could be possible because Tabs permission could be granted to extensions without declaration.

[...]

PR #511, opened 2020-07-19
Remove local file permission

Sneezry:
We added local file permission (`file://*/*`) for https://github.com/Authenticator-Extension/Authenticator/issues/171, and merged the change with https://github.com/Authenticator-Extension/Authenticator/pull/173

This permission is required for some customer use cases (see also https://github.com/Authenticator-Extension/Authenticator/issues/400). From [this answer](https://stackoverflow.com/questions/19493020/adding-file-permission-to-chrome-extension/19493206#19493206) on StackOverflow, the local file permission must be approved by users manually.

However, recently we got [an issue](https://github.com/Authenticator-Extension/Authenticator/issues/510) to report a possible security issue of local file permission. Because Tabs permission could be granted by default:

> You can use most chrome.tabs methods and events without declaring any permissions in the extension's manifest file.
>
> https://developer.chrome.com/extensions/tabs

Extensions with local file permission are able to traverse local files by openning local file urls (for example `file:///C:/`) in new tabs, and read contents with active tab.

This is a serious security issue, and we need to discard `file://*/*` permission ASAP no matter it is disabled by default or not.

For importing data with local image files, we could add a new file input controller to make users be able to upload selected file to the extension as what a regular website does.

### DIFF
--- a/manifest-chrome.json
+++ b/manifest-chrome.json
@@ -52,8 +52,7 @@
     "permissions": [
         "activeTab",
         "storage",
-        "identity",
-        "file://*/*"
+        "identity"
     ],
     "optional_permissions": [
         "clipboardWrite",


mymindstorm:
I don't see this as a major security issue with Authenticator, we are using the `file://*/*` permission as intended. If this is an issue, we should file a bug with Chromium.

> Extensions with local file permission are able to traverse local files by openning local file urls (for example file:///C:/) in new tabs, and read contents with active tab.
>
> This is a serious security issue, and we need to discard file://*/* permission ASAP no matter it is disabled by default or not.

We obviously don't do this, and it would be hard to do file traversal due to activeTab.

- We can only inject a script directly after user interaction with the extension in Chrome (I.e. one click = one attempt at a file). [Docs](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/permissions#activeTab_permission)
- We can't list directory contents, so we would have to know the exact path of the file we would want to read.
- We can't extract data using a content script from files in an unknown or binary format like password databases as they are instantly downloaded and no page is loaded.
- [This would not work in firefox as you cannot open a file path from an extension](https://user-images.githubusercontent.com/27789806/87956616-e2e69200-ca74-11ea-808d-a203fd570a07.png)
- The user still has to enable "Allow access to file URLs" manually. That permission is not granted by default

mymindstorm (approved):

mymindstorm:
You are right, I forgot to take the user's perspective into consideration. A file picker would be better.

Sneezry:
Yeah, a file picker makes more sense. And because Google official authenticator mobile app supports to export data with QR code in recent version, this feature could be our high priority work item.

mymindstorm:
Screenshotting the code is blocked. Opening QR code on phone => taking a picture using webcam => importing codes is a bit of a convoluted process, we could also add an import via webcam page.

Sneezry:
Merge this PR because we are able to import image files with https://github.com/Authenticator-Extension/Authenticator/pull/521 and local file permission is no longer needed.

PR #524, opened 2020-08-12
add pin for entry

Sneezry:


src/store/Accounts.ts
```diff
@@ -20,9 +20,14 @@ export class Accounts implements Module {
       }
     }
 
+    const pinnedEntries = entries.filter(entry => entry.pinned);
+    const unpinnedEntries = entries.filter(entry => !entry.pinned);
```
mymindstorm: These two properties should be getters instead of relying on mutation logic. I don't see entries on inital decrypt.

mymindstorm (changes requested):
If an entry is not pinned, then "pinned" should not be in the backup.

mymindstorm (approved):

PR #526, opened 2020-08-15
add option to hide animated overlay when scanning qr code

jmlag:


mymindstorm:
Thanks for the PR! Why do you want this option, is the overlay too annoying? How could we improve the overlay without adding an option?

jmlag:
Yeah, it makes it hard to see what's on the screen; if you have a QR code in the middle of the screen, the QR code from the animation overlays it which is a bit confusing visually.

I would prefer removing the overlay altogether or replacing it with a text popup that fades away after a few seconds. 

If you don't want to do that, I'd recommend moving it to the upper-left corner, making it smaller, and reducing the opacity.

mymindstorm:
What do you think about an opaque dialog that only shows on first scan with the animation and text? We could put a button at the bottom right to show it again.

jmlag:
Seems like a good solution to me.

nguyendongduy (approved):

mymindstorm:
Closing due to lack of response.

PR #553, opened 2020-11-28
Migrate from Travis CI to GitHub Actions

mymindstorm:
Quick math shows that one of our builds on travis takes about 4 min, meaning that with new travis pricing we can only run 250 builds a month. I want to eventually have integration testing on CI which will eat more minutes, and GH actions should help with that.


Sneezry:
Shall we keep travis while migrating to github actions? And remove travis after github actions work stable.

mymindstorm:
@Sneezry the parts that haven't been tested cannot run in parallel (i18n strings auto-commit, tagging, and auto-release).

mymindstorm:
In case this doesn't work at all we can always just revert to travis.

Sneezry:
Is this PR ready to review?

mymindstorm:
Yes

.github/workflows/i18n.yml
```diff
@@ -0,0 +1,17 @@
+name: i18n
+
+on:
+  push:
+    branches: [ dev ]
+
+jobs:
+  i18n-strings:
+    runs-on: ubuntu-latest
+    name: Process new i18n strings`
```
Sneezry: extra ` in end of line

.github/workflows/main.yml
```diff
@@ -0,0 +1,65 @@
...
+      
+      - name: Install dependencies
+        run: |
+          npm i prettier
+          sudo npm i -g lintspaces-cli
+
+      - name: Prettier
+        run: ./node_modules/prettier/bin-prettier.js --check ./src/* ./src/**/* ./src/**/**/* ./sass/*.scss
+
+      - name: lintspaces
```
Sneezry, line 28: Upper case first letter

.github/workflows/main.yml
```diff
@@ -0,0 +1,65 @@
...
+        run: |
+          lintspaces -nt -d 'spaces' -i 'js-comments' src/*/* src/*.ts view/* manifest-*.json css/popup.css css/import.css
+  build-ffx:
+    runs-on: ubuntu-latest
```
Sneezry, line 32: Do we need set up new environment for build?

.github/workflows/release.yml
```diff
@@ -0,0 +1,46 @@
...
+      
+      - name: Install dependencies
+        run: npm i
+      
+      - name: Build
+        run: bash scripts/release.sh
+        env:
+          CREDS_FILE_PASSWORD: ${{ secrets.CREDS_FILE_PASSWORD }}
+      
+      - name: Create a Release
```
Sneezry: Create a release

Sneezry (changes requested):

.github/workflows/main.yml
mymindstorm: This is intentional

.github/workflows/main.yml
mymindstorm: No, each key under jobs is run in a separate environment.

Sneezry (approved):

Issue #549, opened 2020-11-21
Cant export/add/change settings


Leproide:
**Describe the bug:**
I have activated the encryption, if i decrypt with the correct key the add-on show duplicated, one with the code in clear and one with the request to decrypt.
If I try to export the data it tells me that I have to decrypt first but it has already been decrypted.

**Platform: Windows 10 20H2**
 - Browser: Firefox
 - Browser Version: 83.0 (64 bit)

mymindstorm:
Please follow [these steps](https://authenticator.cc/docs/en/debug-info)

Leproide:
```
Unchecked lastError value: Error: Could not establish connection. Receiving end does not exist. 8 background.js:63
Errore interpretazione XML: non well-formed
Indirizzo: https://userstylesapi.com/tic/stats
Riga numero 1, colonna 1: stats:1:1
Unchecked lastError value: Error: Could not establish connection. Receiving end does not exist. background.js:1
Errore interpretazione XML: non well-formed
Indirizzo: https://userstylesapi.com/tic/stats
Riga numero 1, colonna 1: stats:1:1
Uncaught TypeError: _urlToStyles[id] is undefined
    newStylesLookup moz-extension://9cd3b156-79f5-4c0a-ae90-69a1b5c023fa/src/messaging.js:144
    onload moz-extension://9cd3b156-79f5-4c0a-ae90-69a1b5c023fa/src/messaging.js:324
messaging.js:144:21
Unchecked lastError value: Error: Could not establish connection. Receiving end does not exist. 3 background.js:1
Unchecked lastError value: Error: Could not establish connection. Receiving end does not exist. background.js:63
Unchecked lastError value: Error: Could not establish connection. Receiving end does not exist. 3 background.js:1
Unchecked lastError value: Error: Could not establish connection. Receiving end does not exist. 8 background.js:63
Errore interpretazione XML: non well-formed
Indirizzo: https://userstylesapi.com/tic/stats
Riga numero 1, colonna 1: stats:1:1
Unchecked lastError value: Error: Could not establish connection. Receiving end does not exist. background.js:1
Unchecked lastError value: Error: Could not establish connection. Receiving end does not exist. 2 background.js:63
Errore interpretazione XML: non well-formed
Indirizzo: https://userstylesapi.com/tic/stats
Riga numero 1, colonna 1: stats:1:1
Unchecked lastError value: Error: Could not establish connection. Receiving end does not exist. background.js:63
[...]

PR #554, opened 2020-12-01
Allow export decrypted data only

Sneezry:
Fix for #549

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -126,8 +126,12 @@
         "description": "Update."
     },
     "phrase_incorrect": {
-        "message": "You cannot add a new account or export data until all accounts are decrypted. Please enter the correct password before continuing.",
+        "message": "You cannot add a new account until all accounts are decrypted. Please enter the correct password before continuing.",
         "description": "Passphrase Incorrect."
+    },
+    "phrase_incorrect_export": {
+        "message": "Some accounts are not decrypted, we will skip them in the backup file.",
+        "description": "Skip Unable-decripted Data."
     },
     "phrase_not_match": {
         "message": "Password does not match.",
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -130,7 +130,7 @@
         "description": "Passphrase Incorrect."
     },
     "phrase_incorrect_export": {
-        "message": "Some accounts are not decrypted, we will skip them in the backup file.",
+        "message": "Accounts that were not able to be decrypted will not be included in this backup.",
         "description": "Skip Unable-decripted Data."
     },
     "phrase_not_match": {
--- a/src/components/Popup/ExportPage.vue
+++ b/src/components/Popup/ExportPage.vue
@@ -46,8 +46,7 @@ export default Vue.extend({
       unsupportedAccounts: hasUnsupportedAccounts(exportData),
       exportFile: getBackupFile(exportData),
       exportEncryptedFile: getBackupFile(exportEncData, key),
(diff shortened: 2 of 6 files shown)


_locales/en/messages.json
```diff
@@ -126,9 +126,13 @@
         "description": "Update."
     },
     "phrase_incorrect": {
-        "message": "You cannot add a new account or export data until all accounts are decrypted. Please enter the correct password before continuing.",
+        "message": "You cannot add a new account until all accounts are decrypted. Please enter the correct password before continuing.",
         "description": "Passphrase Incorrect."
     },
+    "phrase_incorrect_export": {
+        "message": "Some accounts are not decrypted, we will skip them in the backup file.",
```
mymindstorm, line 133: "Unencrypted accounts will not be included in the backup file."

mymindstorm: "Accounts that were not able to be decrypted will not be included in this backup."

src/components/Popup/MenuPage.vue
```diff
@@ -124,11 +124,18 @@ export default Vue.extend({
...
+      if (this.$store.getters["accounts/currentlyEncrypted"]) {
+        if (tab === "SetPasswordPage") {
           this.$store.commit("notification/alert", this.i18n.phrase_incorrect);
           return;
         }
+
+        if (tab === "ExportPage") {
+          this.$store.commit(
+            "notification/alert",
+            this.i18n.phrase_incorrect_export
```
mymindstorm: Can you move this warning to ExportPage? Like how we do it with `otp_unsupported_warn` and `export_info`

mymindstorm (changes requested):

_locales/en/messages.json
Sneezry: Updated.

src/components/Popup/MenuPage.vue
Sneezry: Updated.

src/components/Popup/ExportPage.vue
```diff
@@ -43,7 +46,8 @@ export default Vue.extend({
       unsupportedAccounts: hasUnsupportedAccounts(exportData),
       exportFile: getBackupFile(exportData),
       exportEncryptedFile: getBackupFile(exportEncData, key),
-      exportOneLineOtpAuthFile: getOneLineOtpBackupFile(exportData)
+      exportOneLineOtpAuthFile: getOneLineOtpBackupFile(exportData),
+      currentlyEncrypted: this.$store.getters["accounts/currentlyEncrypted"]
```
mymindstorm: Getters should be on `computed`

https://vuex.vuejs.org/guide/getters.html

Sneezry: <del>When user click "Encrypted" and input new password, that value could be changed</del>

You are right, computed executes after data, and varible value won't update once it is initialized. I will update this part.

mymindstorm: <details>
<summary>old comment</summary>

Putting the getter in `computed` will still cause it to update properly, it just looks at the reactive dependencies which will work because of vuex. We are already doing this in other places like filter logic:

https://github.com/Authenticator-Extension/Authenticator/blob/ce0e82962353b6e73dd349294172ace512f00131/src/components/Popup/MainBody.vue#L57-L62

</details>

oops, didn't see that edit

Sneezry: Updated

mymindstorm (approved):

Issue #555, opened 2020-12-04
Import from Google Authenticator official mobile app


Sneezry:
Both iOS and Android version Google Authenticator official mobile app support export data. The export data uses Google private schema of `otpauth-migration:`. Also, the URI is very different from `otpauth:`. Though it's not a standard way, considering to large amount of users of this app, we should support this backup format.

See: https://github.com/google/google-authenticator-android/issues/118

mymindstorm:
closing, implemented

PR #560, opened 2020-12-07
Add otpauth migration support

Sneezry:
#555

### DIFF
--- a/src/components/Import/QrImport.vue
+++ b/src/components/Import/QrImport.vue
@@ -127,7 +127,10 @@ async function getOtpUrlFromQrFile(file: File): Promise<string | null> {
             );
 
             if (jsQrCode && jsQrCode.data) {
-              if (jsQrCode.data.indexOf("otpauth://") !== 0) {
+              if (
+                jsQrCode.data.indexOf("otpauth://") !== 0 &&
+                jsQrCode.data.indexOf("otpauth-migration://") !== 0
+              ) {
                 return resolve(null);
               }
               return resolve(jsQrCode.data);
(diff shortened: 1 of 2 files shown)


mymindstorm (approved):
Looks good, but I haven't had the time to test this just yet. I will open an issue later if I find anything.

mymindstorm (commented):
Works perfectly, although it was not easy to get the QR onto my computer. It's a bit odd that only the import page allows this type of import, can we make this work via normal QR scanning?

Sneezry:
> Works perfectly, although it was not easy to get the QR onto my computer. It's a bit odd that only the import page allows this type of import, can we make this work via normal QR scanning?

The QR code should be captured from mobile App, and transferred to computer. So I think it should be a local image file. Because we have already removed `file://*` permission, we are not able to scan local images via normal QR scanning.

mymindstorm:
> > Works perfectly, although it was not easy to get the QR onto my computer. It's a bit odd that only the import page allows this type of import, can we make this work via normal QR scanning?
> 
> So I think it should be a local image file. Because we have already removed `file://*` permission, we are not able to scan local images via normal QR scanning.

There are some situations where this could happen, e.g. the screenshot is uploaded to Google drive and scanned in from the preview.

Sneezry:
> > > Works perfectly, although it was not easy to get the QR onto my computer. It's a bit odd that only the import page allows this type of import, can we make this work via normal QR scanning?
> > 
> > 
> > So I think it should be a local image file. Because we have already removed `file://*` permission, we are not able to scan local images via normal QR scanning.
> 
> There are some situations where this could happen, e.g. the screenshot is uploaded to Google drive and scanned in from the preview.

Make sense. I will update the PR to add this feature.

Sneezry:
I will make another PR for scanning migration QR code. Closing this one.

PR #567, opened 2020-12-16
Move AddMethodPage to header

mikedidomizio:
Makes it more accessible and not require the user to scroll through
their list to add a new one

# Description

If a user has many codes, it requires them to scroll through the list before they can add another one.  I am one of those users.
This fixes this issue by moving the add button to the header.  I removed the one at the bottom of the list.

## Old way

Mf4D0vEG7Y

## New way

OuVRWEG6Ly

## New way (after clicking button)

GaBHDfKV4I

### DIFF
--- a/sass/popup.scss
+++ b/sass/popup.scss
@@ -66,7 +66,8 @@ svg {
   }
 }
 
-#i-menu {
+#i-menu,
+#i-plus {
   position: absolute;
   left: 20px;
   bottom: 0;
--- a/sass/popup.scss
+++ b/sass/popup.scss
@@ -66,8 +66,7 @@ svg {
   }
 }
 
-#i-menu,
-#i-plus {
+#i-menu {
   position: absolute;
   left: 20px;
   bottom: 0;
@@ -95,6 +94,7 @@ svg {
   }
 }
 
+#i-plus,
 #i-qr {
   position: absolute;
   right: 45px;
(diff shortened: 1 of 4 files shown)


src/components/Popup/MainHeader.vue
```diff
@@ -11,6 +11,15 @@
       >
         <IconCog />
       </div>
+      <div
+        class="icon"
+        id="i-plus"
+        v-bind:title="i18n.settings"
```
mymindstorm: Alt text has the wrong string

sass/popup.scss
```diff
@@ -66,7 +66,8 @@ svg {
   }
 }
 
-#i-menu {
+#i-menu,
+#i-plus {
```
mymindstorm: I think it may be better to align the button to the same spot that the scan button is since the action is similar.

mymindstorm (changes requested):
Thanks for the PR!

sass/popup.scss
mikedidomizio: Updated SASS to put this in the same spot as the scan button

src/components/Popup/MainHeader.vue
mikedidomizio: I've updated the string, I've added `Add code`.

mikedidomizio:
Updated location and alt text

image

mymindstorm (approved):

PR #568, opened 2020-12-17
import migration data via common scanning

Sneezry:


Sneezry:
@mymindstorm is this release triggered as expected? https://github.com/Authenticator-Extension/Authenticator/releases/tag/refs/heads/scan-migration

Sneezry:
There could be an error in our release yaml file. Here's a possible solution: https://github.community/t/how-to-run-github-actions-workflow-only-for-new-tags/16075/27

mymindstorm:
> @mymindstorm is this release triggered as expected? [`refs` (release)](https://github.com/Authenticator-Extension/Authenticator/releases/tag/refs/heads/scan-migration)

Yeah, that's not supposed to happen. I've applied the fix you suggested on dev.

_locales/en/messages.json
```diff
@@ -35,6 +35,18 @@
         "message": "Manual Entry",
         "description": "Manual Entry."
     },
+    "migration_fail": {
+        "message": "Migrate data failed. You need export data from Google Authenticator again.",
+        "description": "Import migration data failed."
+    },
+    "migration_partly_fail": {
+        "message": "Migrate data successfully, but some account data is broken.",
```
mymindstorm:
```suggestion
        "message": "Some account data was not imported successfully."
```

src/background.ts
```diff
@@ -122,17 +124,46 @@ function getQr(
...
-    if (text === "error decoding QR Code") {
-      chrome.tabs.sendMessage(id, { action: "errorqr" });
+    if (text.indexOf("otpauth-migration://") === 0) {
+      const otpUrls = getOTPAuthPerLineFromOPTAuthMigration(text);
+      if (otpUrls.length === 0) {
+        !silent && chrome.tabs.sendMessage(id, { action: "errorenc" });
+        return false;
+      }
+
+      let failedCount = 0;
```
mymindstorm: If we track and notify the user of partial migrations, should we add the same logic to https://github.com/Authenticator-Extension/Authenticator/pull/560/files#diff-baf3e12b03ac667f8d5b47a772b39ac41828fd6745180de5509b37ac0e77721eR224?

_locales/en/messages.json
```diff
@@ -35,6 +35,18 @@
         "message": "Manual Entry",
         "description": "Manual Entry."
     },
+    "migration_fail": {
+        "message": "Migrate data failed. You need export data from Google Authenticator again.",
```
mymindstorm: Do we need all of these new strings? Failed could work here, otherwise use 
```suggestion
        "message": "Import failed. Please re-export your data from Google Authenticator and try again.",
```

_locales/en/messages.json
```diff
@@ -35,6 +35,18 @@
...
+        "description": "Import migration data failed."
+    },
+    "migration_partly_fail": {
+        "message": "Migrate data successfully, but some account data is broken.",
+        "description": "Some migration data is broken."
+    },
+    "migration_success": {
+        "message": "Migrate data successfully.",
```
mymindstorm: `updateSuccess` could work here instead of a new string

_locales/en/messages.json
Sneezry: I prefer the new string, because it guides users the next step.

src/background.ts
```diff
@@ -122,17 +124,46 @@ function getQr(
...
+      for (const otpUrl of otpUrls) {
+        const success = await getTotp(otpUrl, true);
```
mymindstorm: I noticed that it hangs a bit when importing a lot of codes, maybe promise.allSettled these right after the loop?

mymindstorm (approved):
Works great!

src/background.ts
Sneezry: Good suggestion

PR #570, opened 2020-12-19
Fix encodeURIComponent in otpauth URI

emon:
Fixed to properly escape `label` and `issuer` when the extension displays QR codes for otpauth URLs.

For example, if the `issuer` contained `&`, the QR code would not be read properly by Apps on iOS and Android. 
White-spaces and some symbols would cause same problem.


Sneezry:
Hi @emon , could you fix the style issue of this PR?

emon:
Hi. I fixed the style issue.

mymindstorm (approved):
Thanks for the contribution!

PR #574, opened 2020-12-23
Add tests for menu page

mikedidomizio:
## Description

I wanted to understand the project and test setup better and so I chose to add tests for the menu page since it is a pretty simple component.

## Changes

- Adds `sinon-chrome` dependency which allows the mocking of the global chrome object for testing purposes
- Adds basic testing for `MenuPage` component.  I covered all cases of the feedback button except the `feedbackURL` which I don't really understand where it would come from/be set.
- Updated the `test-runner` to allow for `console.log`.  Converting these tests if possible to run in an IDE with breakpoints would be nice.  For those that are unfamiliar with all the inner workings of the project, this will help with debugging purposes when writing tests.
- Updated the `test-runner` to be a bit more type safe.
- `allowSyntheticDefaultImports` for the `tsconfig` since my compiler was complaining about `import chrome from "sinon-chrome";`.


mikedidomizio:
Updating as WIP, the `Test Code` step passed but actually did not run the tests!  Not sure if some flaky error in the test run or some change that I've made at this time.

https://github.com/Authenticator-Extension/Authenticator/pull/574/checks?check_run_id=1600874001

mikedidomizio:
Removed WIP.  I fixed the issue in this commit.  It was an issue with using `--single-process` that I introduced.

[Fix commit here](https://github.com/Authenticator-Extension/Authenticator/pull/574/commits/2c8e6d90226191c1a2edad59a03c213de7c27189)

Although I've included it in this pull request.  I'm fine with creating a separate pull request to resolve this issue if need be.

mymindstorm:
> Although I've included it in this pull request. I'm fine with creating a separate pull request to resolve this issue if need be.

That's fine, I've added changes to make the script handle errors properly. Please ignore the code coverage errors, I need to fork `@jsdevtools/coverage-istanbul-loader` to generate a basic coverage map.

>  I covered all cases of the feedback button except the feedbackURL which I don't really understand where it would come from/be set.

This is set via `chrome.storage.managed` which is controlled via GPO or a file placed on your computer by a sysadmin. More info [here](https://authenticator.cc/docs/en/policy-config) and [here](https://github.com/Authenticator-Extension/Authenticator/blob/1d123a01645af4b4257f7a16d8c4d5a52ecb477f/src/store/Menu.ts#L20). You should be able to just mock `chrome.storage.managed` before the component is initialized to test it.

> if possible to run in an IDE with breakpoints 

That should be possible by getting creative with https://marketplace.visualstudio.com/items?itemName=msjsdiag.debugger-for-chrome

mikedidomizio:
I took a crack at adding a test for the `feedbackURL` path.  So things get complicated with the amount of mocking that has to occur before the `feedbackURL` is set and making everything work properly.  Also depending on what you mock and what you don't mock, as some things have dependencies on others.

For instance in [`popup.ts`](https://github.com/Authenticator-Extension/Authenticator/blob/dev/src/popup.ts#L36), there's all the set up of the vuex modules.  If you mock out too much, the Vuex instance doesn't properly get created.

Anyways, I'm attempting to write a test for it but it's a bit tricky, at least with the route I'm trying.

mymindstorm:
It seems possible, albeit not exactly clean looking.

https://gist.github.com/mymindstorm/9cd1d474f3edc70b19e4ee65a25861e7

mikedidomizio:
> It seems possible, albeit not exactly clean looking.
> 
> https://gist.github.com/mymindstorm/9cd1d474f3edc70b19e4ee65a25861e7

This is where I was heading except I kept going deeper and deeper.  So in terms of clean looking, it's definitely cleaner than what I had.

I've committed your patch and pushed it onto this pull request.

PR #591, opened 2021-02-08
Notice user to use import local file instead of scanning local files

Sneezry:
Because we have deprecated `file://` permission in the recent version, and some user use scanning local file to import backups heavily, we need popup a notification and guide the user to use import tool for local QR images.

Snipaste_2021-02-08_13-40-08

Snipaste_2021-02-08_13-40-29

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -193,6 +193,10 @@
         "message": "Capture failed, please reload the page and try again.",
         "description": "Capture Failed"
     },
+    "capture_local_file_failed": {
+        "message": "Are you trying to scan QR code from a local file? Use Import QR Image Backup instead.",
+        "description": "Import QR image backup instead of scan local image"
+    },
     "based_on_time": {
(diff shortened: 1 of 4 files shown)


Sneezry:
addons-linter@2.18.0 has an issue to validate linter. The real error message is

```
{
  ruleId: null,
  fatal: true,
  severity: 2,
  message: 'Parsing error: Invalid ecmaVersion.',
  line: undefined,
  column: undefined
}
```

addons-linter throws `JS_SYNTAX_ERROR` when `message.fatal` is true, but `JS_SYNTAX_ERROR` and fatal error are two different things.

The root cause of this issue is addons-linter uses 2021 as the default ecmaVersion, however, the validator is not able to recognize this version. This is an issue of addons-linter.

The build from the latest code of addons-linter repository works as expected with no errors.

Snipaste_2021-02-13_05-38-01

mymindstorm (approved):
Looks good!

Issue #633, opened 2021-03-24
OneDrive: 'account loading...'


jbostoen:
**Describe the bug:** just after rolling out this extension for more users, it seems there's an issue between Authenticator and OneDrive. The backup page is stuck on 'account - loading...'

Perhaps something changed on OneDrive, or perhaps it's a bug; how do we debug this?

**Platform:** Windows 10
 - Browser: Chrome, FireFox
 - Browser Version: latest versions

In Azure's audit logs, we do see a lot of "app role assignment grant to user" failures (Microsoft.Online.DirectoryServices.UniqueKeyPropertyException)

With my personal/private OneDrive account, it works as expected; but not within our corporate O365 environment.

---

After I explicitly sign out and then sign in again, this appears in the debugger:

Error 403 - XHR GET https://graph.microsoft.com/v1.0/me/drive/special/approot

Request

```
Accept-Encoding | gzip, deflate, br
-- | --
Accept-Language | en-US,en;q=0.5
Authorization | Bearer   <redacted>
Connection | keep-alive
Cookie | s_preRef=%5B%5B%27neowin.net%27%2C%271492531998951%27%5D%5D;  s_fid=1DB9DB9943B5A8EF-1329F236115FC0EE;  _ga=GA1.2.1340837329.1525429298; LPVID=BkZWMyMWJiOTZjMzdhZWE5;  mbox=PC#b4abd76be1114733b279653e440d54b7.37_0#1677239875\|session#e6381b68c4914f58b5d0e3e2fa11f6ce#1613996935;   MC1=GUID=e539776adb3b46ccb03fb00329f2a55d&HASH=e539&LV=202002&V=4&LU=1580719930235;  MUID=246626B83BF46FDE02132CD03FF46C26; MSCC=1614337854;  mslocale={'u':'nl-be'}
Host | graph.microsoft.com
```

Response
```
{
  "error": {
    "code": "accessDenied",
    "message": "Access denied",
    "innerError": {
      "date": "2021-03-24T10:08:39",
      "request-id": "b65d8e4a-064f-4465-a4ac-0d8ec86b1206",
      "client-request-id": "b65d8e4a-064f-4465-a4ac-0d8ec86b1206"
    }
  }
}
````

Potentially related:
* https://github.com/OneDrive/onedrive-api-docs/issues/682
[...]

PR #638, opened 2021-03-31
Support for OneDrive Business

jbostoen:
* Initial work on an attempt to support OneDrive Business
* Also took the liberty to correct a Dutch translation and update one untranslated string
* fixes a typo in the OneDrive (personal) approach

As discussed in #633 : OneDrive is stuck on 'account loading...' and does not work properly for professional accounts.

I noticed that you just added a new comment with a different approach, but this is what I had now and which seems to be working: it adds a separate page for OneDrive Business.

Only thing to consider: use the "common" string in the authorization URL; or go for "organizations" or even offer the option for a tenant ID. For now (and to avoid potential conflicts with OneDrive personal), I just went with "common".

This work might still help you for the working version :)

### DIFF
--- a/_locales/nl/messages.json
+++ b/_locales/nl/messages.json
@@ -375,12 +375,12 @@
         "message": "Algoritme"
     },
     "smart_filter": {
-        "message": "Slim filter"
+        "message": "Slimme filter"
     },
     "backup": {
         "message": "Backup"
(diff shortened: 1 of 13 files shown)


mymindstorm:
Closing in favor of #641

PR #641, opened 2021-04-03
Onedrive Business account permissions fix #633

### LINKED ISSUE #633: OneDrive: 'account loading...'

mymindstorm:


jbostoen:
Thank you very very much for integrating this!

Should there be a note about permissions to the organization's administrators? Or perhaps to the user telling them they might need to consult their admin?

mymindstorm:
> Should there be a note about permissions to the organization's administrators?

Added

> Or perhaps to the user telling them they might need to consult their admin?

The consent screen will prompt users to request admin approval if they cannot authorize the app. At some point in the future, the consent screen will mark the integration as "untrustworthy" due us not being able to register it with a verified publisher (no legal entity to register as a MS Partner with).

image

src/background.ts
```diff
@@ -331,7 +331,9 @@ function getBackupToken(service: string) {
       redirUrl = encodeURIComponent(chrome.identity.getRedirectURL());
       authUrl = `https://login.microsoftonline.com/common/oauth2/v2.0/authorize?client_id=${
         getCredentials().onedrive.client_id
-      }&response_type=code&redirect_uri=${redirUrl}&scope=https%3A%2F%2Fgraph.microsoft.com%2FFiles.ReadWrite.AppFolder%20https%3A%2F%2Fgraph.microsoft.com%2FUser.Read%20offline_access&response_mode=query&prompt=consent`;
+      }&response_type=code&redirect_uri=${redirUrl}&scope=https%3A%2F%2Fgraph.microsoft.com%2FFiles.ReadWrite${
+        localStorage.oneDriveBusiness === "false" ? ".AppFolder" : ""
```
Sneezry: Shoud `localStorage.oneDriveBusiness !== "true"` make more sense here?

mymindstorm: They both work, I just thought that this would be easier to read.

Sneezry: My concern is if `localStorage.oneDriveBusiness` is `undefined`, this logic should be incorrect.

jbostoen: I'm unfamiliar with extensions or the settings, but I was mostly wondering why there are quotes around the boolean?

Sneezry: localStorage can only store string values

Sneezry (approved):

jbostoen:
Any idea when this will result in a new release? 😁

mymindstorm:
See #646

Sneezry:
> Any idea when this will result in a new release? 😁

For Chrome, because we have more than 1.5M users, we need incremental rollouts. It could be as long as one month to deliver this version to all existing users. New-install users are able to install this version once it's available in CWS. This version should be available in CWS within one week.

For Firefox, @mymindstorm correct me if I have any mistake, it should be already available to all users, including existing users and new-install users.

For Edge, we may need some time before it is available to public, but it could not be too long.

jbostoen:
Thanks @Sneezry !

Question: is the 'manual sync' actually a one way backup rather than a sync?

I set up the Authenticator with a OneDrive Business on 2 different computers. 
They both already had different TOTPs; in the JSON file on OneDrive Business I only see a backup of one of those 2 lists at a time.

mymindstorm:
> For Firefox, @mymindstorm correct me if I have any mistake, it should be already available to all users, including existing users and new-install users.

It is usually immediately available. They _used_ to have incremental rollouts but removed it for some reason.

> For Edge, we may need some time before it is available to public, but it could not be too long.


Edge usually takes a week or two for certification and is available immediately. 


> Question: is the 'manual sync' actually a one way backup rather than a sync

We will have to correct the string. It is a one way backup and is referenced that way in all other parts of the ui. The only "sync" is the browsers built in extension data sync. I would be wary of using it without backups because it tends to be finicky.

jbostoen:
Hm, thanks for the answer @mymindstorm . I'll have to discuss it with colleagues, but it it would be more clear to remove the word "sync" as it implies two-way sync

PR #648, opened 2021-04-22
fix(onedrive): fix onedrive permission doc link unclickable

Sneezry:


mymindstorm (approved):
That is weird. It works just fine in ffox. I really need to get around to writing all those unit tests so we don't have to deal with these problems.

sedabasfirinci:
633527

PR #654, opened 2021-04-29
Added password field to prepend to otp when copying to clipboard

cesaregiannetti:
Some providers use to concatenate the password and the otp, this makes it more convenient.
This work was done for myself, quick and dirty, but with your help might become useful to others.

Schermata 2021-04-29 alle 16 39 22
Schermata 2021-04-29 alle 16 39 56

### DIFF
--- a/_locales/ar/messages.json
+++ b/_locales/ar/messages.json
@@ -389,5 +389,8 @@
     },
     "backup_file_info": {
         "message": "Backup your data to a file."
+    },
+    "password": {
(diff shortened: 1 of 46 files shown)


Sneezry:
This will make 2-factor authentication useless, because you put the 2 factors in one same place.

cesaregiannetti:
> This will make 2-factor authentication useless, because you put the 2 factors in one same place.

Good point.. well, I tried

PR #660, opened 2021-05-09
Fix 3rd party backup errors

### LINKED ISSUE #659: Importing data does not work
<More info here: https://authenticator.cc/docs/en/lost-codes>

**Describe the bug:** When I import my data from the extension, it doesn't work. I tired importing the JSON file and the text file from my previous backup, when I clicked the button nothing happened. No warning, no progress, nothing. I tired reinstalling the extension to no avail.

**Platform:** <This is required>
 - Browser: Edge
 - Browser Version: Version 90.0.818.56 (Official build) (64-bit)

mymindstorm:
fixes #659

### DIFF
--- a/src/components/Import/FileImport.vue
+++ b/src/components/Import/FileImport.vue
@@ -65,6 +65,7 @@ export default Vue.extend({
               (key) => ["key", "enc", "hash"].indexOf(key) === -1
             ).length;
           } catch (e) {
+            console.warn(e);
             const result = await getEntryDataFromOTPAuthPerLine(
               reader.result as string
             );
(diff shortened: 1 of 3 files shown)


src/components/Import/FileImport.vue
```diff
@@ -75,8 +76,10 @@ export default Vue.extend({
 
           let key: { enc: string; hash: string } | null = null;
 
-          if (importData.key) {
-            key = importData.key;
+          if (importData.hasOwnProperty("key")) {
```
Sneezry, line 79: Is it necessary? `importData` is parsed from a JSON string, and not an instance of a class. `importData` has no inheritance from the prototype.

mymindstorm: Yes, the error with backup creates data like this:
```json
{
    "key": null
}
```
Because of that
```js
Boolean(importData.key) == false
```
so the only way to check if the key exists if it could be null is:
```js
importData.hasOwnProperty("key")
// or
"key" in importData
```
I am not aware of any relation between hasOwnProperty and classes, I am under the impression that it is a method for objects.

Sneezry: I see.

Sneezry (approved):

PR #689, opened 2021-06-21
add 1..9 hotkeys

dutzi:
This PR adds a keyboard shortcut to copy one of the codes.

Open the popup, hit 1, 2, ... 9 to copy the first, second, etc... codes that are listed.

### DIFF
--- a/src/popup.ts
+++ b/src/popup.ts
@@ -147,6 +147,41 @@ async function init() {
     false
   );
 
+  document.body.addEventListener("keydown", (e) => {
+    if (
+      document.activeElement?.nodeName === "INPUT" ||
+      document.activeElement?.nodeName === "TEXTAREA"
+    ) {
+      return;
+    }
+
+    const index = Number(e.key);
+    if (Number.isNaN(index)) {
+      return;
+    }
+
+    const numEntries = instance.$store.state.accounts.entries.length;
+
+    if (index > numEntries) {
+      return;
+    }
+
+    const codeClipboard = document.querySelector<HTMLInputElement>(
+      "#codeClipboard"
+    );
+
+    if (!codeClipboard) {
+      return;
+    }
+
+    codeClipboard.value =
+      instance.$store.state.accounts.entries[index - 1].code;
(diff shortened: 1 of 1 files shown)


src/popup.ts
```diff
@@ -147,6 +147,41 @@ async function init() {
...
+    const codeClipboard = document.querySelector<HTMLInputElement>(
+      "#codeClipboard"
+    );
+
+    if (!codeClipboard) {
+      return;
+    }
+
+    codeClipboard.value =
+      instance.$store.state.accounts.entries[index - 1].code;
```
Sneezry, line 178: index here could be zero, and `index - 1` could be `-1`

Sneezry: `instance.$store.state.accounts.entries` could not be shown by sort of the index in UI. Some accounts may be pinned, some accounts may be filtered.
Here's how are entries rendered: https://github.com/Authenticator-Extension/Authenticator/blob/dev/src/components/Popup/MainBody.vue#L26-L40

dutzi:
That’s true.

dutzi:
👍🏻

src/popup.ts
```diff
@@ -147,6 +147,41 @@ async function init() {
     false
   );
 
+  document.body.addEventListener("keydown", (e) => {
+    if (
+      document.activeElement?.nodeName === "INPUT" ||
+      document.activeElement?.nodeName === "TEXTAREA"
```
Sneezry, line 153: I'm not sure if there's a better way to handle this. But obviously, editable `div` is not included, and read-only input, read-only textarea, radio and checkbox are not excluded.

dutzi:
Right. Maybe stopPropagation on each input?

Sneezry:
> Right. Maybe stopPropagation on each input?

The sample code is not awesome, but it works: https://jsbin.com/vevovuyefu/edit?html,js,console,output

Sneezry:
Closing in favor of #691 

Feel free to reopen this PR if you believe it is closed by mistake.

PR #691, opened 2021-06-24
add key binding to copy code

Sneezry:
Press arrow keys to select a code, press `enter` to copy.

Animation

### DIFF
--- a/src/components/Popup/MainBody.vue
+++ b/src/components/Popup/MainBody.vue
@@ -14,6 +14,7 @@
         v-model="searchText"
         v-bind:placeholder="i18n.search"
         type="text"
+        tabindex="-1"
       />
       <div id="searchHint" v-if="searchText === ''">
         <div></div>
@@ -25,10 +26,10 @@
     <div
       v-dragula
       drake="entryDrake"
-      v-on:keydown.down="focusNextEntry(entries, shouldFilter, filter)"
-      v-on:keydown.right="focusNextEntry(entries, shouldFilter, filter)"
-      v-on:keydown.up="focusLastEntry(entries, shouldFilter, filter)"
-      v-on:keydown.left="focusLastEntry(entries, shouldFilter, filter)"
+      v-on:keydown.down="focusNextEntry()"
+      v-on:keydown.right="focusNextEntry()"
+      v-on:keydown.up="focusLastEntry()"
+      v-on:keydown.left="focusLastEntry()"
     >
       <EntryComponent
         v-for="entry in entries"
@@ -36,7 +37,7 @@
         v-bind:filtered="!entry.pinned && !isMatchedEntry(entry)"
         v-bind:notSearched="!isSearchedEntry(entry)"
         v-bind:entry="entry"
-        v-bind:tabindex="getTabindex(entry, entries, shouldFilter, filter)"
+        v-bind:tabindex="getTabindex(entry)"
       />
     </div>
   </div>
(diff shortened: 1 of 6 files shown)


src/components/Popup/MainBody.vue
```diff
@@ -90,6 +90,87 @@ export default Vue.extend({
...
     },
+    isEntryVisible(entry: OTPEntry, shouldFilter: boolean, filter: boolean) {
+      return (
+        this.isSearchedEntry(entry) &&
+        (entry.pinned || !shouldFilter || !filter || this.isMatchedEntry(entry))
+      );
+    },
+    getTabindex(
+      entry: OTPEntry,
+      entries: OTPEntry[],
```
mymindstorm, line 101: entries, shouldFilter, and filter are all computed properties and don't need to be passed as arguments.

src/components/Popup/MainBody.vue
```diff
@@ -90,6 +90,87 @@ export default Vue.extend({
...
+    ) {
+      const firstEntry = entries.find((entry) =>
+        this.isEntryVisible(entry, shouldFilter, filter)
+      );
+
+      return entry === firstEntry ? 0 : -1;
+    },
+    findNextEntryIndex(
+      entries: OTPEntry[],
+      shouldFilter: boolean,
```
mymindstorm, line 113: same as above.

src/components/Popup/MainBody.vue
```diff
@@ -90,6 +90,87 @@ export default Vue.extend({
...
+      filter: boolean
+    ) {
+      const firstEntry = entries.find((entry) =>
+        this.isEntryVisible(entry, shouldFilter, filter)
+      );
+
+      return entry === firstEntry ? 0 : -1;
+    },
+    findNextEntryIndex(
+      entries: OTPEntry[],
```
mymindstorm: On all functions, entries, shouldFilter, and filter are all computed properties and don't need to be passed as arguments.

mymindstorm:
I really like this!

Bugs / Other:
- When entering password, it does not function. Need to hit tab or reopen the extension for it to work.
- If I click anywhere else (even an entry) it will not let me use arrow keys until tab or reopen. 
	- Should we listen for keys on the document?
	- Should we focus entries on click?

src/components/Popup/MainBody.vue
Sneezry: Updated.

Sneezry:
> When entering password, it does not function. Need to hit tab or reopen the extension for it to work.

Fixed. After applying password, the first entry will be auto focused.

> If I click anywhere else (even an entry) it will not let me use arrow keys until tab or reopen.

Entry will have focus when it is clicked. If you click anywhere else, the entry should not have focus. And is as designed.

> Should we listen for keys on the document?

I don't think so. By following w3 accessibility guide, `tab` should be used to select active element, and arrow keys should be used to select items in an active group element. If the group element is not active, we should not response to the action.

> Should we focus entries on click?

Yes. This is a bug in previous commit. I have fixed that in the recent commit.

src/components/Popup/MainBody.vue
```diff
@@ -14,6 +14,7 @@
         v-model="searchText"
         v-bind:placeholder="i18n.search"
         type="text"
+        tabindex="-1"
```
mymindstorm, line 17: I think this should be able to be tabbed to, but it is ok to not allow it until we get around to overhaling keyboard nav for the header and menu. Hopefully compoentizing everything will help at the second version of that.

mymindstorm (approved):
Looks great

mymindstorm:
Will this supersede #689?

Sneezry:
> Will this supersede #689?

I suppose. Number keys could be hard to use to copy a specific entry. Imaging how to copy the 7th entry in the list. I have to count entries from top to bottom.

Issue #676, opened 2021-05-24
Sync function with google drive not working


dangdan2807:
**Describe the bug:**
After adding the new 2fa code, I clicked backup on select google after waiting a few minutes but still didn't see the success message but normally and when I go to Google Drive > Authenticator Backups I still don't see the new backup .json file. (connected to google account)

**Platform:**
 - Browser: Chorme
 - Browser Version: 90.0.4430.212

PR #693, opened 2021-06-30
Add GOST 34.11 hash algorithm

dmitrykatenko:
Add OTP generation using GOST 34.11 hash function with 256/512 bit length (gost.ts, key-utilities.ts). Add necessary options and enumerations (AddAccountPge.vue, otp.ts).


src/models/key-utilities.ts
```diff
@@ -91,6 +92,28 @@ export class KeyUtilities {
...
   }
 
+  static CryptoJsWordArrayToUint8Array(wordArray: CryptoJS.lib.WordArray) {
+    const l = wordArray.sigBytes;
+    const words = wordArray.words;
+    const result = new Uint8Array(l);
+    let i = 0 /*dst*/,
+      j = 0; /*src*/
+    // eslint-disable-next-line no-constant-condition
+    while (true) {
```
Sneezry: Seems `while(i < l)` makes more sense here?

src/models/key-utilities.ts
```diff
@@ -91,6 +92,28 @@ export class KeyUtilities {
...
+      // here i is a multiple of 4
+      if (i == l) break;
```
Sneezry: Should always use `===` instead of `==`

src/models/key-utilities.ts
dmitrykatenko: Agree with you. Corrected

src/models/key-utilities.ts
dmitrykatenko: OK, made a commit

Sneezry (approved):
LGTM

mymindstorm (commented):
My preliminary concerns before looking at the code too closely / testing:
- I have not heard of this algorithm before, is it used for TOTP? It appears to be nonstandard as it is not referenced in RFC 6238
- Due to the above I would like to see automated tests on this. It will be harder to maintain nonstandard algorithms from our end, so tests would be very beneficial. I will be able to help with this if needed when I have some free time.
- It looks like you took code from another one of your company's projects to implement GOST. Is this code available as a library on npm?

dmitrykatenko:
- the totp algorithm allows a variety of hash functions to be used on the client and provider, including GOST. Since the hash function is also an input parameter of the totp algorithm. http://tools.ietf.org/html/rfc6986 - RFC to hash function.
- ok, we will work on it
- it is not in the npm repository, but it's probably not a problem to place

dmitrykatenko:
Implemented npm package usage

dmitrykatenko:
tests added

mymindstorm (changes requested):
Although the test you have added is good to verify that the library is functioning, what I am looking for is code that preforms a test with known good HOTP secrets and codes against codes generated with GOST from `KeyUtilites.generate`. I want to make sure that any future changes to our code does not impact GOST since we will not be testing it manually. After that we can merge this.

dmitrykatenko:
made a test for predefined secret and date with known/valid otp code

mymindstorm (approved):

mymindstorm:
Thanks for the changes!

dmitrykatenko:
please tell me when the release is planned with our changes?

mymindstorm:
We don't have anything planned, but I have been looking to fix #676 before the next release.

PR #702, opened 2021-07-08
fix base32 bug

### LINKED ISSUE #701: This plugin does not work with Google App
I use it for 2 different systems, but I added a new one today, while 2 of them are correct, the 3th generates incorrect code, the phone application and the code produced by the plugin do not match

image

 - Browser: Chrome
 - Browser Version: 91.0.4472.124

Sneezry:
Fix #701 

The root cause of this bug is we check `low` value not equals to zero to ditermine the concat data is not pad in base32.

This logic is incorrect. If the real data's low part is zeros in real, the computed base32 will miss an `A` at the end.

### DIFF
--- a/src/models/migration.ts
+++ b/src/models/migration.ts
@@ -6,8 +6,10 @@ function byteArray2Base32(bytes: number[]) {
   let result = "";
   let high = 0,
     low = 0,
-    sh = 0;
+    sh = 0,
+    hasDataInLow = false;
   for (let i = 0; i < len; i += 5) {
+    hasDataInLow = true;
     high = 0xf8 & bytes[i];
     result += chars.charAt(high >> 3);
     low = 0x07 & bytes[i];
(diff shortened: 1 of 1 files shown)


mymindstorm (approved):
I honestly don't fully understand this part of the code, trusting that you have resolved this issue. Although, I think that this would be a good place to throw some hard coded tests.

Sneezry:
Here's `Illustrations and Examples` of base32 RFC to help you to understand the logic if you would like to look deeply for this part of code: https://datatracker.ietf.org/doc/html/rfc4648#section-9

mymindstorm:
> Here's `Illustrations and Examples` of base32 RFC to help you to understand the logic if you would like to look deeply for this part of code: [datatracker.ietf.org/doc/html/rfc4648#section-9](https://datatracker.ietf.org/doc/html/rfc4648#section-9)

Thanks, that document is useful.

PR #706, opened 2021-07-09
fix #705

### LINKED ISSUE #705: Scan QR code, get "You cannot add an account until all accounts are decrypted"
Consider:
 * a clean Google Chrome [91] browser with a fresh installation of Authenticator.cc [6.3.2]
 * a pre-installed policy for this app which contains "enforcePassword: true"

When you go to add an MFA for the first time, Authenticator will ask you to set a password.  (A side issue is that this password doesn't seem to be remembered if there are no stored MFAs - you can end up having to set the password several times if the process aborts, or if you remove the last MFA from the app - but that's not the subject of this bug report although it may be part of the underlying issue.)

So on a fresh installation, this sequence of events happens:
 * press the QR icon at the top of the app
 * Get asked to set a password
 * Press OK, get a box that says "Success", but the QR scanner doesn't open.
 * Press the QR icon again
 * Scan the QR code
 * get a dialogue which says "You cannot add a new account until all accounts are decrypted.  Please enter the correct password before continuing."

Thus it appears to be impossible to add a first MFA using the QR code.

It works well if the data is entered manually; and once the first MFA has been added, it is then possible to add subsequent MFAs using the QR code.

Sneezry:
fix #705

### DIFF
--- a/src/background.ts
+++ b/src/background.ts
@@ -269,8 +269,8 @@ async function getTotp(text: string, silent = false) {
           entryData[hash].algorithm = algorithm;
         }
         if (
-          (await EntryStorage.hasEncryptedEntry()) !==
-          encryption.getEncryptionStatus()
+          (await EntryStorage.hasEncryptedEntry()) &&
+          !encryption.getEncryptionStatus()
         ) {
           !silent && chrome.tabs.sendMessage(id, { action: "errorenc" });
           return false;


mymindstorm (commented):
This is more of a larger issue with how we are checking for encrypted entries rather than just this conditional. With the key encryption, all of these encryption checks should be looking for if there is a key present rather than an encrypted entry. I will look into writing some code for this.

Sneezry:
> This is more of a larger issue with how we are checking for encrypted entries rather than just this conditional. With the key encryption, all of these encryption checks should be looking for if there is a key present rather than an encrypted entry.

I agree. When there is no entries, and a password is set, current logic will just ignore the key.

mymindstorm:
Closing in favor of #724

PR #707, opened 2021-07-09
fix gost package issue

### LINKED ISSUE #700: External dependency crypto-gost tries to inject script to background page
TL;DR:

crypto-gost lib tries to load `gostEngine.js` without Worker mode.

image

Detail:

crypto-gost needs its own engine component locates inside the module folder. But there seems a code bug in this lib, and the worker check never works.

image

The condition used in L6464 is always `false`, and `worker` is always `undefined` so that the lib always try to inject `gostEngine.js` with HTML tag.

This lib is hosted in a private repo, and we cannot report bugs: https://github.com/vostokplatform/crypto/issues

Sneezry:
fix #700

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,9919 @@
 {
   "name": "authenticator-extension",
   "version": "0.1.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "name": "authenticator-extension",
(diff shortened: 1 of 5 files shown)


dmitrykatenko:
Changes looks good.
OTP Valid.

Sneezry:
> Changes looks good.
> OTP Valid.

Thanks for the review and validation😊

mymindstorm (approved):

PR #724, opened 2021-08-19
Fix various bugs

mymindstorm:


src/background.ts
```diff
@@ -269,7 +269,8 @@ async function getTotp(text: string, silent = false) {
           entryData[hash].algorithm = algorithm;
         }
         if (
-          (await EntryStorage.hasEncryptedEntry()) !==
+          // If the entries are encrypted and we aren't unlocked, error.
```
Sneezry, line 272: What if the entries are not encrypted, but we are unlocked? I think #705 is in such situation.

Sneezry (approved):

PR #727, opened 2021-08-21
refactor ManagedStorage for save get

Sneezry:
The hard coded `ManagedStorage.get` default value `false` doesn't make sense. We also have keys with string values, and their default values should not be `false`.

### DIFF
--- a/src/models/storage.ts
+++ b/src/models/storage.ts
@@ -4,7 +4,9 @@ import * as uuid from "uuid/v4";
 
 export class BrowserStorage {
   private static async getStorageLocation() {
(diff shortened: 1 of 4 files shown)


src/models/storage.ts
```diff
@@ -4,7 +4,9 @@ import * as uuid from "uuid/v4";
 
 export class BrowserStorage {
   private static async getStorageLocation() {
-    const managedLocation = await ManagedStorage.get("storageArea");
+    const managedLocation = await ManagedStorage.get<"sync" | "local">(
```
mymindstorm, line 7: Shouldn't this be string?

mymindstorm (approved):

PR #729, opened 2021-08-27
add no entry message

Sneezry:
image


_locales/en/messages.json
```diff
@@ -431,5 +431,8 @@
     },
     "learn_more": {
         "message": "Learn more"
+    },
+    "no_entires": {
+        "message": "No entries to display. Add your first account now."
```
mymindstorm: We need to make our terminology here consistent to avoid confusion. `entries` should be `accounts` in all strings.

mymindstorm (commented):
This looks pretty good!

_locales/en/messages.json
Sneezry: Updated.

mymindstorm (approved):

PR #731, opened 2021-09-02
add context menu support

Sneezry:


manifests/manifest-firefox.json
```diff
@@ -47,6 +47,7 @@
     "optional_permissions": [
         "<all_urls>",
         "clipboardWrite",
+        "contextMenus",
```
mymindstorm: This is not an allowed optional permission in Firefox. I think it may be best to just remove Firefox support for this feature rather than adding a required permission.

src/background.ts
```diff
@@ -590,3 +592,47 @@ async function setAutolock() {
...
+    (result) => {
+      if (result) {
+        if (localStorage.enableContextMenu === "true") {
+          chrome.contextMenus.create({
+            title: chrome.i18n.getMessage("extName"),
+            contexts: ["all"],
+            onclick: (_, tab) => {
+              let popupUrl = "view/popup.html?popup=true";
+              if (tab.url && tab.title) {
+                popupUrl +=
```
mymindstorm, line 610: Inconsistent behavior with filtering on example.com:

Screenshot from 2021-09-04 13-31-41

src/store/Accounts.ts
```diff
@@ -497,20 +497,33 @@ export class Accounts implements Module {
...
+        }
 
-        if (!tab.url) {
+        if (!url) {
           return resolve([title, null]);
         }
 
         const urlParser = document.createElement("a");
-        urlParser.href = tab.url;
+        urlParser.href = url;
```
mymindstorm: We should convert this to [URL.hostname](https://developer.mozilla.org/en-US/docs/Web/API/URL/hostname) rather than messing with the DOM.

mymindstorm (changes requested):

manifests/manifest-firefox.json
Sneezry: Updated

src/background.ts
Sneezry: Updated

image

src/store/Accounts.ts
Sneezry: Updated

mymindstorm (approved):

PR #827, opened 2022-01-01
add permissions management

Sneezry:
image


src/components/Permissions.vue
```diff
@@ -0,0 +1,32 @@
...
+    <h1>Permissions</h1>
+    <div v-for="permission in permissions" :key="permission.id">
+      <h2>{{ permission.id }}</h2>
+      <p>{{ permission.description }}</p>
+      <button
+        :disabled="!permission.revocable"
+        v-if="permission.revocable"
+        v-on:click="revoke(permission.id)"
+      >
+        Revoke
```
mymindstorm: String not in messages.json

src/store/Permissions.ts
```diff
@@ -0,0 +1,246 @@
+import { Permission } from "../models/permission";
+
+const permissions: Permission[] = [
+  {
+    id: "activeTab",
```
mymindstorm, line 5: I would suggest removing `storage` and `identity` because they are not mentioned in the Chrome or Firefox UI nor actually give any meaningful access.

`activeTab` is also already mentioned in the Chrome UI as part of the extension dropdown menu.

Screenshot from 2022-01-02 12-29-15


If you would rather keep the non-revocable permissions in this menu, I think it would be useful to add a message saying "Can not revoke permission" instead of leaving it blank.

Sneezry: Add an option to show required permissions

image

There's also a message to say the required permission cannot be revoked

image

mymindstorm (approved):

PR #831, opened 2022-01-06
add autofill command

Sneezry:
https://github.com/Authenticator-Extension/Authenticator/discussions/829

### DIFF
--- a/src/background.ts
+++ b/src/background.ts
@@ -579,10 +579,6 @@ chrome.commands.onCommand.addListener(async (command: string) => {
         }
       );
 
-      if (cachedPassphrase === "") {
-        return;
-      }
-
       chrome.tabs.query(
         { active: true, lastFocusedWindow: true },
         async (tabs) => {
--- a/src/background.ts
+++ b/src/background.ts
@@ -8,6 +8,8 @@ import { Encryption } from "./models/encryption";
 import { EntryStorage, ManagedStorage } from "./models/storage";
 import { Dropbox, Drive, OneDrive } from "./models/backup";
 import * as uuid from "uuid/v4";
+import { getSiteName, getMatchedEntries } from "./utils";
+import { CodeState } from "./models/otp";
 
 import { getOTPAuthPerLineFromOPTAuthMigration } from "./models/migration";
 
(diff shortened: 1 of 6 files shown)


mymindstorm:
So, the reason we didn't implement this before was because we don't have secure website matching like the password managers do.  We can't be sure that a website will not intentionally trick our auto fill matching and pretend to be a legitimate website. Do you think that is a big enough issue to not implement this feature?

Sneezry:
> So, the reason we didn't implement this before was because we don't have secure website matching like the password managers do. We can't be sure that a website will not intentionally trick our auto fill matching and pretend to be a legitimate website. Do you think that is a big enough issue to not implement this feature?

The current design is if there is only one matched account, autofill command will fill the code, or nothing happens. I don't think it's a big problem because in most cases, there's only one matched account. Smart filter works well, and secure website matching is not so important. FYI, this PR is for this feedback https://github.com/Authenticator-Extension/Authenticator/discussions/829

philippthiele:
I am a little confused, my authenticator extension already has such an option, I enabled it to use autofill, but it doesn't do anything. Now I see here it is not really merged yet. What does the current autofill option do?

EDIT:
Oh I see now, I had to restart chrome and now when I click the code it autofills. Sorry for the too quick question. But I will be very interested to see this feature here merged, I think it is a good addition, since I already thought this is what it is doing with the current autofill.

src/background.ts
```diff
@@ -560,6 +562,59 @@ chrome.commands.onCommand.addListener(async (command: string) => {
...
+          if (!tab || !tab.id) {
+            return;
+          }
+          contentTab = tab;
+
+          const siteName = await getSiteName();
+          const entries = await EntryStorage.get();
+          const matchedEntries = getMatchedEntries(siteName, entries);
+
+          if (matchedEntries && matchedEntries.length === 1) {
```
mymindstorm, line 595: matchedEntries.length <= 1

src/background.ts
```diff
@@ -560,6 +562,59 @@ chrome.commands.onCommand.addListener(async (command: string) => {
...
+              }
+            );
+          } catch (error) {
+            console.error(error);
+            return reject(error);
+          }
+        }
+      );
+
+      if (cachedPassphrase === "") {
```
mymindstorm: This can be removed since you account for unencrypted entries.

src/background.ts
Sneezry: if `matchedEntries.length == 0`, `const entry = matchedEntries[0]` will be `undefined`. We should auto fill code when there is only one matched entry

mymindstorm: Sorry, incorrect logic on my end. I meant `matchedEntries.length >= 1`. It won't past a code if there is more than one match.

Sneezry: If there is more than one matched entry I think we should always ask the user to choose the correct one, but not the first one. What do you think?

mymindstorm: I agree, that would be the best solution.

src/background.ts
Sneezry: If the extension is locked, and the user presses the hotkey to auto-fill code, this code block can prevent the command. Do I understand correctly?

mymindstorm: Yes, but doesn't this also prevent unencrypted access?

Sneezry: Oh, I got your point. Let me update this.

Sneezry: Updated

Haiyangha:
gh pr checkout 831

Haiyangha:
gh pr checkout 831

src/background.ts
Jeunesse16: Birden fazla eşleşen giriş varsa, kullanıcıdan her zaman doğru olanı seçmesini istememiz gerektiğini düşünüyorum, ancak ilkini değil. Ne düşünüyorsun?

Jeunesse16:
Birden fazla eşleşen giriş varsa, kullanıcıdan her zaman doğru olanı seçmesini istememiz gerektiğini düşünüyorum, ancak ilkini değil. Ne düşünüyorsun?

PR #907, opened 2022-05-23
Add 'wasm-unsafe-eval' to custom CSP (fix #906)

### LINKED ISSUE #906: "Error: Please use $(ref:runtime.getURL)." in latest Firefox Nightly
### Describe the issue

After entering the password to decrypt data, a "Loading..." prompt appears but nothing happens.
In the browser console, the message "Error: Please use $(ref:runtime.getURL)." is shown (moz-extension://4e8dab5e-4299-4207-a974-f6037c40ea76/dist/popup.js:22).

### Browser

Firefox

### Browser Version

102.0a1

### Extension Version

6.3.4

mozfreddyb:
It seems that some dependencies (at least argon2-browser) require allowing WebAssembly at the CSP level.
Recent Firefox allows that by default, but having your own CSP means this extension won't be able to benefit from this automatically.

### DIFF
--- a/manifests/manifest-firefox.json
+++ b/manifests/manifest-firefox.json
@@ -57,5 +57,5 @@
         "https://graph.microsoft.com/me/*",
         "https://login.microsoftonline.com/common/oauth2/v2.0/token"
     ],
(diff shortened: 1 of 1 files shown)


mozfreddyb:
FWIW, I do not intend to do more work on this pull request. Feel free to steal / adjust and merge at your own convenient. :)

mymindstorm:
Thanks for this! Although, it seems that MV2 extensions are not affected by this just yet per bug 1770909

mozfreddyb:
Indeed, we changed course after we learned that it would impact too many add-ons.

PR #924, opened 2022-07-04
Fix broken locales

rebornix:
As title. Some locale files are not valid JSON files, which can not be loaded by chromium properly. Might be introduced via merge conflict resolution.


Sneezry:
@mymindstorm are these files generated by crowdin?

mymindstorm:
Yes, I'll have to look at it.

mymindstorm:
Closing, fixed by b2d1701. This might have been caused by the script at https://github.com/Authenticator-Extension/Authenticator/blob/dev/scripts/i18n.sh

PR #927, opened 2022-07-10
Use more restrictive CSP in Chromium

mymindstorm:
We may want to wait a while or save this for the MV3 transition due to the minimum version requirement.

https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Content_Security_Policy#webassembly

### DIFF
--- a/manifests/manifest-chrome.json
+++ b/manifests/manifest-chrome.json
@@ -50,7 +50,7 @@
         "pages": [
             "view/argon.html"
         ],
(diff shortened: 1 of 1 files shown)


Sneezry (approved):

Sneezry:
Do we still need this PR?

mymindstorm:
Yes, but we can delay it.

Issue #1048, opened 2023-05-10
disableBackup doesn't work


RedDeer3:
### Describe the issue

Hello,

we are deploying this in an enterprise and using the policies outlined [here](https://authenticator.cc/docs/en/policy-config)

Edge has been configured here: Policies\Microsoft\Edge\3rdparty\extensions\ocglkepbibnalbgmbachknglpdipeoio\policy

The policies below are enabled:

Working:
passwordPolicy
passwordPolicyHint
enforcePassword
disableExport

Not Working
disableBackup

Set as "disableBackup"="1"

Policies are set with [HKEY_LOCAL_MACHINE\SOFTWARE\Policies\Microsoft\Edge\3rdparty\extensions\ocglkepbibnalbgmbachknglpdipeoio\policy]

Thanks

### Browser

Edge

### Browser Version

Version 113.0.1774.3

### Extension Version

6.3.3.3

Sneezry:
Hi RedDeer3, I have a quick test and everything works on my side:

Without disableBackup:

![image](https://github.com/Authenticator-Extension/Authenticator/assets/1621293/4cc15ef1-abd6-4262-a2bd-e66ca05b0eaa)

With disableBackup:

![image](https://github.com/Authenticator-Extension/Authenticator/assets/1621293/664580a9-b1bd-4067-ae20-5607cf0e0e39)

After setting disableBackup to 1, all backup download buttons are gone

Sneezry:
@RedDeer3 thanks for the feedback! I can repro this issue on my side. A fix is on the way.

Sneezry:
We've got a solution for the issue: a fixed release for it can be found at https://github.com/Authenticator-Extension/Authenticator/releases/tag/v6.3.5. However, there's a slight hiccup: Microsoft currently doesn't accept any MV2 extension packages for Edge, which means we're unable to release the fix to the Edge store just yet. We're working on the MV3, but it's still a work in progress. In the meantime, we've already rolled out the fixed version to CWS where it's currently undergoing review, without being automatically rejected for MV2 reasons.

[...]

PR #1050, opened 2023-05-10
Fix disable backup policy not working issue

Sneezry:
See https://github.com/Authenticator-Extension/Authenticator/issues/1048


Sneezry:
@mymindstorm this is an urgent fix, we should fix it asap

mymindstorm:
Assuming that you tested this, then it looks good to me.

mymindstorm (approved):

PR #1221, opened 2024-07-01
Implement work-around/fix for incorrect render of manually added HOTP tokens

hicallmeal:
The parsed `<option>` returns the OTP.type as a string, instead of the required integer type. Later in execution, the code can't handle the string in comparisons, etc. 

As a result, selecting a different type causes the resulting entry to be handled as a standard TOTP type. Giving it the appearance and behaviour as such - it also affects backups (..`&counter=undefined`) if done so in the same session. 
Any type of operation apart from the default type (as that's assigned an integer by default) with the option is affected. 
image

In saying that, once reloaded, some other part of the code corrects the entry and storage. 

This just parseInts the string so that it's correctly handled.
You might have some better way of doing this, perhaps earlier on, etc.

### DIFF
--- a/src/components/Popup/AddAccountPage.vue
+++ b/src/components/Popup/AddAccountPage.vue
@@ -33,7 +33,7 @@
         <option :value="OTPAlgorithm.GOST3411_2012_256">GOST 34.11 256</option>
         <option :value="OTPAlgorithm.GOST3411_2012_512">GOST 34.11 512</option>
       </a-select-input>
-      <a-select-input :label="i18n.type" v-model="newAccount.type">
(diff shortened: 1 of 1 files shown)


mymindstorm (approved):
Good catch!

mymindstorm:
Could you fix the CI issue and manually test the changes? `npm run chrome` -> `chrome://extensions` should allow you to do so.

hicallmeal:
I've updated the patch according to commit notes; essentially in order to return an integer, a `.number` modifier is required. 
https://v2.vuejs.org/v2/guide/forms.html#number

Have tested and compiled successfully.

mymindstorm (approved):

PR #1238, opened 2024-07-21
Create SECURITY.md

mymindstorm:


SECURITY.md
```diff
@@ -0,0 +1,11 @@
+# Security Policy
+
+## Supported Versions
+
+We support any version published via our official distribution channels.
```
Sneezry: I suggest that we only ensure that the Chrome and Firefox versions can be fixed as soon as possible, because Edge and Safari have relatively large API changes and may have technical limitations.

SECURITY.md
```diff
@@ -0,0 +1,11 @@
...
+
+## Reporting a Vulnerability
+
+Report potential vulnerabilities privately via [this form](https://github.com/Authenticator-Extension/Authenticator/security/advisories/new).
+Where appropriate, include a proof-of-concept and reproduction steps.
+We will provide an initial response within 5 days.
```
Sneezry: For a purely free open source project, there is a lot of pressure to respond to any security report within 5 days. In fact, for non-urgent issues or issues that we do not think must be fixed, we can delay the response. During the holidays, when it is not convenient to work, it is very likely that we will not be able to respond within 5 days.

SECURITY.md
mymindstorm: Before the big update a few years ago Edge was very different, but now it is almost the same as Chrome. I might just be forgetting, but what are the big differences in Edge? Given that we have 2M+ active users there, we should support it. For Safari I agree with you, unless @rebornix would like to add.

SECURITY.md
mymindstorm: "inital response" here means just acknowledging that we have received the issue. I can change it to "as soon as reasonably possible for the maintainers"?

SECURITY.md
mymindstorm: If we move forward with not supporting safari, we should indicate it is a "community version" or similar in the README.

SECURITY.md
Sneezry: How do you think about "We will strive to provide an initial response within 5 days, though it is not guaranteed."

mymindstorm: That sounds appropriate to me.

PR #1246, opened 2024-07-29
Require verifying password before chaning it

mymindstorm:
I plan to make another PR to deduplicate password verification and hashing code.


src/components/Popup/SetPasswordPage.vue
```diff
@@ -80,6 +104,19 @@ export default Vue.extend({
...
 
       this.$store.commit("currentView/changeView", "LoadingPage");
+
+      if (this.defaultEncryption) {
+        const isCorrectPassword = await verifyPasswordUsingKeyID(
+          this.defaultEncryption,
+          this.currentPhrase
+        );
+        if (!isCorrectPassword) {
+          this.$store.commit("notification/alert", this.i18n.phrase_not_match);
```
Sneezry: We should distinguish between the situations where the user has entered an incorrect current password and where the two new passwords do not match, to avoid causing confusion for the user.

PR #1255, opened 2024-08-15
i18n -- MiD fixes

mymindstorm:


mymindstorm:
there are a few issues -- fixing...

PR #1264, opened 2024-08-21
fix add account issue

### LINKED ISSUE #1262: Adding new accounts with manual entry is broken
### Describe the issue

while doing manual entry it's not showing the code 

### Browser

Chrome

### Browser Version

127.0.6533.120

### Extension Version

8.0.0

Sneezry:
Fix https://github.com/Authenticator-Extension/Authenticator/issues/1262

### DIFF
--- a/src/components/Popup/AddAccountPage.vue
+++ b/src/components/Popup/AddAccountPage.vue
@@ -115,6 +115,11 @@ export default Vue.extend({
         this.newAccount.period = undefined;
       }
 
+      const defaultEncyptionKey = this.$store.state.accounts.defaultEncryption;
+      const encryption = this.$store.state.accounts.encryption[
+        defaultEncyptionKey
+      ];
+
       const entry = new OTPEntry(
         {
           type,
@@ -128,7 +133,7 @@ export default Vue.extend({
           digits: this.newAccount.digits,
           algorithm: this.newAccount.algorithm,
         },
-        this.$store.state.accounts.encryption
+        encryption
       );
 
       await entry.create();


src/models/otp.ts
```diff
@@ -88,9 +88,21 @@ export class OTPEntry implements OTPEntryInterface {
           hash: string;
           index: number;
         },
-    encryption?: EncryptionInterface
+    encryption?: EncryptionInterface | Map<string, EncryptionInterface>
```
mymindstorm, line 91: This should just be a `Map<string, EncryptionInterface>`, the only place we need to have a case for `EncryptionInterface` is before migration logic runs.

src/models/otp.ts
```diff
@@ -88,9 +88,21 @@ export class OTPEntry implements OTPEntryInterface {
...
   ) {
-    this.encryption = encryption;
+    if (encryption instanceof Map) {
+      for (const key in encryption) {
+        if (
+          /^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$/.test(
+            key
+          )
+        ) {
+          this.encryption = encryption.get(key);
```
mymindstorm, line 100: Use `state.accounts.defaultEncryption` instead of a loop

mymindstorm (changes requested):
Please ignore previous comments.

Instead of making OTPEntry have multiple encryptions (we need to maintain a 1:1 relationship between an OTPEntry and an encryption), make 

https://github.com/Authenticator-Extension/Authenticator/blob/dev/src/components/Popup/AddAccountPage.vue#L131 

pass `state.accounts.defaultEncryption` instead.

Sneezry:
Updated

mymindstorm (approved):

PR #1283, opened 2024-08-26
Fix OTP type

### LINKED ISSUE #1271: Can't set period to 60
### Describe the issue

If I try to import a URI otpauth://totp/test:test?secret=XXXXX&period=60&digits=6&issuer=work, the period of 60 seconds is never used. I can confirm this because when I export a backup, I just get otpauth://totp/test:test?secret=XXXXX&issuer=work

### Browser

Chrome

### Browser Version

127.0.6533.89

### Extension Version

8.0.0

Sneezry:
Fix https://github.com/Authenticator-Extension/Authenticator/issues/1271

We have both OTPType and string definitions for OTP entry types. Previously, OTPType was a numeric enum. When importing OTP URLs, OTP entries were generated with a numeric type value, causing an issue where the period parameter was ignored. This happened because we were comparing OTPType.totp (which has a numeric value of 1) with the string "totp", leading to a false condition. As a result, the system incorrectly identified the entry as not time-based, causing the period parameter to be disregarded. This fix ensures all OTP types are now aligned with the new OTPType string enum, resolving the issue.

### DIFF
--- a/src/models/otp.ts
+++ b/src/models/otp.ts
@@ -3,12 +3,12 @@ import { UserSettings } from "./settings";
 import { EntryStorage } from "./storage";
 
 export enum OTPType {
-  totp = 1,
-  hotp,
-  battle,
-  steam,
-  hex,
-  hhex,
+  totp = "totp",
+  hotp = "hotp",
+  battle = "battle",
+  steam = "steam",
+  hex = "hex",
+  hhex = "hhex",
 }
 
 export enum CodeState {
@@ -223,8 +223,7 @@ export class OTPEntry implements OTPEntryInterface {
     this.period = decryptedData.period || 30;
     this.pinned = decryptedData.pinned || false;
     this.secret = decryptedData.secret;
-    // @ts-expect-error need a better way to do this
-    this.type = OTPType[decryptedData.type] || OTPType.totp;
+    this.type = decryptedData.type || OTPType.totp;
 
     if (this.type !== OTPType.hotp && this.type !== OTPType.hhex) {
       this.generate();
--- a/src/models/otp.ts
+++ b/src/models/otp.ts
@@ -117,7 +117,30 @@ export class OTPEntry implements OTPEntryInterface {
       this.secret = entry.secret;
     }
 
-    this.type = entry.type;
+    // We may have already had some error OTP type entries in the storage
+    // totp = 1
+    // hotp = 2
+    // battle = 3
+    // steam = 4
+    // hex = 5
+    // hhex = 6
+
+    if ((entry.type as unknown) === 1) {
+      this.type = OTPType.totp;
+    } else if ((entry.type as unknown) === 2) {
+      this.type = OTPType.hotp;
+    } else if ((entry.type as unknown) === 3) {
+      this.type = OTPType.battle;
+    } else if ((entry.type as unknown) === 4) {
+      this.type = OTPType.steam;
+    } else if ((entry.type as unknown) === 5) {
+      this.type = OTPType.hex;
+    } else if ((entry.type as unknown) === 6) {
+      this.type = OTPType.hhex;
+    } else {
+      this.type = entry.type;
+    }
+
     if (entry.issuer) {
       this.issuer = entry.issuer;
     } else {
@@ -224,6 +247,30 @@ export class OTPEntry implements OTPEntryInterface {
     this.pinned = decryptedData.pinned || false;
     this.secret = decryptedData.secret;
     this.type = decryptedData.type || OTPType.totp;
+
+    // We may have already had some error OTP type entries in the storage
+    // totp = 1
+    // hotp = 2
+    // battle = 3
+    // steam = 4
+    // hex = 5
+    // hhex = 6
+
+    if ((decryptedData.type as unknown) === 1) {
+      this.type = OTPType.totp;
+    } else if ((decryptedData.type as unknown) === 2) {
+      this.type = OTPType.hotp;
+    } else if ((decryptedData.type as unknown) === 3) {
+      this.type = OTPType.battle;
+    } else if ((decryptedData.type as unknown) === 4) {
+      this.type = OTPType.steam;
+    } else if ((decryptedData.type as unknown) === 5) {
+      this.type = OTPType.hex;
+    } else if ((decryptedData.type as unknown) === 6) {
+      this.type = OTPType.hhex;
+    } else {
+      this.type = OTPType.totp;
+    }
 
     if (this.type !== OTPType.hotp && this.type !== OTPType.hhex) {
       this.generate();
--- a/src/models/storage.ts
+++ b/src/models/storage.ts
@@ -630,8 +630,32 @@ export class EntryStorage {
         default:
           // we need correct the type here
           // and save it
-          type = OTPType.totp;
-          entryData.type = OTPType.totp;
+
+          // We may have already had some error OTP type entries in the storage
+          // totp = 1
+          // hotp = 2
+          // battle = 3
+          // steam = 4
+          // hex = 5
+          // hhex = 6
+
+          if ((entryData.type as unknown) === 1) {
+            type = OTPType.totp;
+          } else if ((entryData.type as unknown) === 2) {
+            type = OTPType.hotp;
+          } else if ((entryData.type as unknown) === 3) {
+            type = OTPType.battle;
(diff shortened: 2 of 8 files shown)


mymindstorm:
I feel like there are other places where this change needs to be made. E.g. https://github.com/Authenticator-Extension/Authenticator/blob/c8a651e52b597a16d2182d7c7b7844ef549375a1/src/definitions/module-interface.d.ts#L65

We really need to spend some time writing E2E tests. This change is scary to me because this enum is used in a lot of places where type checking is half broken due to vuex.

Sneezry:
Also fixes: https://github.com/Authenticator-Extension/Authenticator/issues/1302 https://github.com/Authenticator-Extension/Authenticator/issues/1292 https://github.com/Authenticator-Extension/Authenticator/issues/1291

@mymindstorm we should have a hotfix ASAP.

mymindstorm:
found a bug:

1. add a HOTP code
2. set an encryption password

the hotp code becomes a totp code

mymindstorm:
this also occurs if you add a hotp when encryption is on and reload the extension

src/models/otp.ts
```diff
@@ -223,8 +246,31 @@ export class OTPEntry implements OTPEntryInterface {
...
+      this.type = OTPType.hotp;
+    } else if ((decryptedData.type as unknown) === 3) {
+      this.type = OTPType.battle;
+    } else if ((decryptedData.type as unknown) === 4) {
+      this.type = OTPType.steam;
+    } else if ((decryptedData.type as unknown) === 5) {
+      this.type = OTPType.hex;
+    } else if ((decryptedData.type as unknown) === 6) {
+      this.type = OTPType.hhex;
+    } else {
```
mymindstorm, line 271: This line is causing the bug. This if statement is not considering if `decryptedData.type` is a valid `OTPType` already and will always return `OTPType.totp` when `decryptedData.type` is a string.

Recommend making a type guard for `OTPType` and using it whenever we need to check if a `type` is valid, such as in `storage.ts` and elsewhere. Plus another shared function to convert from legacy otptype to new version. That would help prevent subtle mistakes like this and help remove all this casting to unknown.

mymindstorm (changes requested):
This has a bug in it.

src/models/storage.ts
```diff
@@ -617,29 +614,53 @@ export class EntryStorage {
       }
 
       if (!entryData.type) {
-        entryData.type = OTPType[OTPType.totp];
+        entryData.type = OTPType.totp;
```
mymindstorm, line 617: move this into the case statement or otherwise refactor so we don't have to cast to unknown below. we can set `type: unknown` in `otp.d.ts`/`RawOTPStorage`

src/models/storage.ts
```diff
@@ -617,29 +614,53 @@ export class EntryStorage {
...
+            type = OTPType.battle;
+          } else if ((entryData.type as unknown) === 4) {
+            type = OTPType.steam;
+          } else if ((entryData.type as unknown) === 5) {
+            type = OTPType.hex;
+          } else if ((entryData.type as unknown) === 6) {
+            type = OTPType.hhex;
+          } else {
+            type = OTPType.totp;
+          }
```
olfek, line 656: I recommend keeping numeric enums where you get the benefit of reverse mappings so you don't have to do this ugly thing.

src/models/otp.ts
```diff
@@ -223,8 +246,31 @@ export class OTPEntry implements OTPEntryInterface {
...
+      this.type = OTPType.totp;
+    }
```
olfek, line 274: (2) I recommend keeping numeric enums where you get the benefit of reverse mappings so you don't have to do this ugly thing.

src/models/otp.ts
```diff
@@ -117,7 +117,30 @@ export class OTPEntry implements OTPEntryInterface {
...
-    this.type = entry.type;
+    // We may have already had some error OTP type entries in the storage
+    // totp = 1
+    // hotp = 2
+    // battle = 3
+    // steam = 4
+    // hex = 5
+    // hhex = 6
+
+    if ((entry.type as unknown) === 1) {
```
olfek, line 128: Entries are stored with a **string** `type`
https://github.com/Authenticator-Extension/Authenticator/blob/6511920bac98576e3c97fe2d213b23e9dc899422/src/models/storage.ts#L225
so why is this `if` block needed?

src/models/otp.ts
```diff
@@ -223,8 +246,31 @@ export class OTPEntry implements OTPEntryInterface {
...
+
+    // We may have already had some error OTP type entries in the storage
+    // totp = 1
+    // hotp = 2
+    // battle = 3
+    // steam = 4
+    // hex = 5
+    // hhex = 6
+
+    if ((decryptedData.type as unknown) === 1) {
```
olfek, line 259: (2) Entries are stored with a **string** `type`
https://github.com/Authenticator-Extension/Authenticator/blob/6511920bac98576e3c97fe2d213b23e9dc899422/src/models/storage.ts#L225
so why is this `if` block needed?

src/models/otp.ts
olfek: And if you were thinking `entry.type` might be a **string** number such as `"1"`, the `===` comparison won't allow it.

I'm not saying you're wrong, just trying to understand it 🙂

olfek:
Flow in question:

1. `src\background.ts > getTotp()`
2. `src\models\storage.ts > EntryStorage > import()`
3. `src\models\otp.ts > OTPEntry`
4. `src\models\storage.ts > EntryStorage > getOTPStorageFromEntry()`
5. `src\models\storage.ts > EntryStorage > get()`

So in the currently deployed extension we have this line:

https://github.com/Authenticator-Extension/Authenticator/blob/6511920bac98576e3c97fe2d213b23e9dc899422/src/models/storage.ts#L481

`(parseInt(data[hash].type) as OTPType)` is always `NaN` because `data[hash].type` is a TEXT string.

`OTPType[OTPType.totp]` gives the string `totp` but the type of `type` is the enum `OTPType`.

From this point onwards, all equality checks on `type` fail and default `type` with relevant code is always in use. For example:

https://github.com/Authenticator-Extension/Authenticator/blob/6511920bac98576e3c97fe2d213b23e9dc899422/src/models/otp.ts#L156-L160

Input `period` is lost and entry is saved with the default value of `30`. Other properties may be affected too.

https://github.com/Authenticator-Extension/Authenticator/blob/6511920bac98576e3c97fe2d213b23e9dc899422/src/models/storage.ts#L221-L227

Type of `type` is string, `entry.type` is a string, `OTPType[entry.type]` gives a number, a number is stored.

https://github.com/Authenticator-Extension/Authenticator/blob/6511920bac98576e3c97fe2d213b23e9dc899422/src/models/storage.ts#L624-L638

Now when loading an entry back in, the number `type` doesn't match any string `type` so the `switch` goes to `default`.

**Pretty significant bug I think, this PR is a high priority one for sure.**

src/models/otp.ts
olfek: Error OTP type entries would only ever be `1`, there is no need to check the other numbers.

[See my comment below](https://github.com/Authenticator-Extension/Authenticator/pull/1283#issuecomment-2378100678)

src/models/otp.ts
olfek: (2) Error OTP type entries would only ever be `1`, there is no need to check the other numbers.

[See my comment below](https://github.com/Authenticator-Extension/Authenticator/pull/1283#issuecomment-2378100678)

olfek:
The cause of this type bug:

https://github.com/Authenticator-Extension/Authenticator/blob/6511920bac98576e3c97fe2d213b23e9dc899422/src/models/storage.ts#L481

https://www.typescriptlang.org/play/?#code/FAehAIGUDsEsAd4FMAu4CMxhOgVwLbgBiA9icAN7Dg3gBCAggErAC+WAxidAM5ooBPZAC5iZcAF5wACngBDAE48kASWgppAImgk0c8HnwAjJAs0BKcHJ5iSlgD73bAbVIkAdIyYBdANxYwKDhEVHAAJiwcAlsIqlp6ZkkMNk5uPnBBZDDRNzCk2UVlNQ1tXSsDAhMzS2sYhydc1zIwz2Y-AIhAXg3AVD2gA

Do you think this is a TypeScript bug? @mymindstorm @Sneezry ?

PR #1009, opened 2023-01-31
Migrate to MV3 (Chrome, Firefox, Edge)

Sneezry:
Fixes #565

- [x] Move to `chrome.storage.session` to cache password
- [x] Move to `chrome.scripting` to inject scan QR script
- [x] Replace `content_security_policy` with `content_security_policy.extension_pages` or `content_security_policy.sandbox` as appropriate.
- [x] Replace `setTimeout` with `chrome.alarms` API for auto-lock
- [x] Replace `XMLHttpRequest` with `fetch` in backend agent
- [x] Migrate LocalStorage to `chrome.storage.local`
- [x] <del>Migrate `Image` and `canvas` to `OffscreenCanvas`(?) for service worker use</del> Move QR decode to content script
- [ ] E2E test
  + [ ] Chrome 
	  + [x] Can create new accounts
	  + [x] Can create new accounts by scanning QR codes
	  + [x] Can import existing accounts
	  + [x] Can migrate data with existing installed MV2 extension
	  + [x] Can set password
	  + [x] Can click to copy
	  + [x] Can click to auto fill
	  + [ ] Can auto backup to 3rd party storage
	  + [x] Auto-lock works as expected
	  + [x] Force lock works as expected
	  + [x] Context menu works as expected
	  + [x] Policy configuration works as expected
	  + [x] Storage area migration works as expected
	  + [ ] TBD
  + [ ] Firefox
	  + [x] Can create new accounts
	  + [x] Can create new accounts by scanning QR codes
	  + [x] Can import existing accounts
	  + [x] Can migrate data with existing installed MV2 extension
	  + [x] Can set password
	  + [x] Can click to copy
	  + [x] Can click to auto fill
	  + [x] Can auto backup to 3rd party storage
	  + [x] Auto-lock works as expected
	  + [x] Force lock works as expected
	  + [ ] ~Context menu works as expected~
	  + [ ] Policy configuration works as expected
	  + [x] Storage area migration works as expected
	  + [ ] TBD
  + [ ] Edge
	  + [x] Can create new accounts
	  + [x] Can create new accounts by scanning QR codes
	  + [x] Can import existing accounts
	  + [x] Can migrate data with existing installed MV2 extension
	  + [x] Can set password
	  + [x] Can click to copy
	  + [x] Can click to auto fill
	  + [ ] Can auto backup to 3rd party storage
	  + [x] Auto-lock works as expected
	  + [x] Force lock works as expected
	  + [x] Context menu works as expected
[...]

PR #1288, opened 2024-09-01
Fix toolbar button shortcut (#1274)

### LINKED ISSUE #1274: Cannot set toolbar hotkey
### Describe the issue

I tried setting toolbar hotkey but it no longer works. Same hotkeys works for other addons. This applies to all functions (Activate toolbar, Scan QR, Autofill).

### Browser

Firefox

### Browser Version

129.0

### Extension Version

8.0.0

cesar19004:
Fix #1274 Since manifest v3 `execute_browser_action` is no longer available, use the `execute_action` command instead
https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/commands

### DIFF
--- a/manifests/manifest-firefox.json
+++ b/manifests/manifest-firefox.json
@@ -28,7 +28,7 @@
     "scripts": ["dist/background.js"]
   },
   "commands": {
-    "_execute_browser_action": {},
+    "_execute_action": {},
     "scan-qr": {
       "description": "Scan a QR code"
     },


Sneezry:
@mymindstorm I think I missed this migration for Firefox in #1009 . Could you help to double confirm? Thanks.

mymindstorm (approved):
lgtm

PR #1310, opened 2024-09-17
[FIX] Seconds not properly prevented from being negative

olfek:
The current code ...

https://github.com/Authenticator-Extension/Authenticator/blob/6511920bac98576e3c97fe2d213b23e9dc899422/src/store/Accounts.ts#L90

... fails in these two examples:

```
0 - 61 + 60 < 0
59 - 120 + 60 < 0
```

The offset (middle value) can be as low as `-300`.

https://github.com/Authenticator-Extension/Authenticator/blob/6511920bac98576e3c97fe2d213b23e9dc899422/src/syncTime.ts#L27

So the least you can have is `-241 (59 - 300)`, the `+ 60` is not enough to make a positive number.

---

Without this fix, in the examples given above, the effect will be:

Countdown circle animation may be out of sync with the current time step because `animationDelay` will be stuck at `0s`.

https://github.com/Authenticator-Extension/Authenticator/blob/6511920bac98576e3c97fe2d213b23e9dc899422/src/components/Popup/EntryComponent.vue#L22-L29

`timeout` class will never activate unless `entry.period < 5` because the left side of the `<` condition will be stuck at `entry.period`.

https://github.com/Authenticator-Extension/Authenticator/blob/6511920bac98576e3c97fe2d213b23e9dc899422/src/components/Popup/EntryComponent.vue#L59

### DIFF
--- a/src/store/Accounts.ts
+++ b/src/store/Accounts.ts
@@ -90,10 +90,10 @@ export class Accounts implements Module {
           }
 
           if (second < 0) {
-            // Handle the situation where offset causes `second` to be negative. #1310 
-            second = 60 - ((second * -1) % 60)
+            // Handle the situation where offset causes `second` to be negative. #1310
+            second = 60 - ((second * -1) % 60);
           } else {
-            second = second % 60
+            second = second % 60;
           }
           state.second = second;
 
--- a/src/store/Accounts.ts
+++ b/src/store/Accounts.ts
@@ -89,8 +89,12 @@ export class Accounts implements Module {
             second += Number(UserSettings.items.offset);
           }
 
-          // prevent second from negative
-          second = (second < 0) ? (60 - (second * -1) % 60) : (second % 60);
+          if (second < 0) {
+            // Handle the situation where offset causes `second` to be negative. #1310 
+            second = 60 - ((second * -1) % 60)
+          } else {
+            second = second % 60
+          }
           state.second = second;
 
           let currentlyEncrypted = false;
--- a/src/store/Accounts.ts
+++ b/src/store/Accounts.ts
@@ -90,7 +90,7 @@ export class Accounts implements Module {
             second += Number(UserSettings.items.offset) + 60;
           }
 
-          second = second % 60;
+          second = (second < 0) ? (60 - (second * -1) % 60) : (second % 60);
           state.second = second;
 
           let currentlyEncrypted = false;
--- a/src/store/Accounts.ts
+++ b/src/store/Accounts.ts
@@ -87,7 +87,7 @@ export class Accounts implements Module {
           let second = new Date().getSeconds();
           if (UserSettings.items.offset) {
             // prevent second from negative
-            second += Number(UserSettings.items.offset) + 60;
+            second += Number(UserSettings.items.offset);
           }
 
           second = (second < 0) ? (60 - (second * -1) % 60) : (second % 60);
--- a/src/store/Accounts.ts
+++ b/src/store/Accounts.ts
@@ -86,10 +86,10 @@ export class Accounts implements Module {
         updateCodes(state: AccountsState) {
           let second = new Date().getSeconds();
           if (UserSettings.items.offset) {
-            // prevent second from negative
             second += Number(UserSettings.items.offset);
           }
 
+          // prevent second from negative
           second = (second < 0) ? (60 - (second * -1) % 60) : (second % 60);
           state.second = second;
 
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,1 @@
+*.disabled


olfek:
Requesting review @Sneezry @mymindstorm

mymindstorm:
It's on my to-do list, but might take me a while to get to this

olfek:
🛎️🛎️🛎️ @Sneezry @mymindstorm

olfek:
@mymindstorm 📟📟📟

src/store/Accounts.ts
```diff
@@ -86,11 +86,11 @@ export class Accounts implements Module {
...
           let second = new Date().getSeconds();
           if (UserSettings.items.offset) {
-            // prevent second from negative
-            second += Number(UserSettings.items.offset) + 60;
+            second += Number(UserSettings.items.offset);
           }
 
-          second = second % 60;
+          // prevent second from negative
+          second = (second < 0) ? (60 - (second * -1) % 60) : (second % 60);
```
mymindstorm: Just to make things a bit more obvious, lets make the OOO clearer and remove the ternary operator to show what the comment is referring to. 

```suggestion
          if (second < 0) {
            // Handle the situation where offset causes `second` to be negative. #1310 
            second = 60 - ((second * -1) % 60)
          } else {
            second = second % 60
          }
```

mymindstorm (commented):
Thanks for your patience. Slight nitpicks, but otherwise good catch!

src/store/Accounts.ts
olfek: hi and thanks for the suggestion, you can go ahead and make these changes to speed this up as I'm currently unavailable 🙂

olfek: oh wait, I can accept the changes via the web interface, which I can't do in the mobile app 🙄

mymindstorm (approved):

mymindstorm:
It looks like I don't have permission to push to your branch to fix the lint issues.

image

olfek:
@mymindstorm I've added you as a collaborator.

olfek:
FYI, I've never been a fan of prettier, it's too opinionated. The actual code is important, how it's formatted can always be corrected later.

olfek:
@mymindstorm I think we're good to go here

olfek:
@mymindstorm ...

After running `prettier` with the `--write` option, I get this from `git diff --stat` :

```
 src/store/Accounts.ts | 79 +++++++++++++++++++++++++--------------------------
 1 file changed, 38 insertions(+), 41 deletions(-)
```

Would you like me to commit this? - It will pollute the PR diff a little, I await your decision.

mymindstorm:
That doesn't seem right, have you tried running `npm run chrome`? It will automatically format it using our configs.

olfek:
@mymindstorm You have collaborator access to my fork, can you take over from here?

olfek:
@mymindstorm ...

> That doesn't seem right, have you tried running `npm run chrome`? It will automatically format it using our configs.

I don't think you have any prettier configs (https://prettier.io/docs/en/configuration)

mymindstorm:
When you use prettier outright instead of through npm run, depending on how your system is configured it might use a different version which may have different rules.

mymindstorm:
I'll get to this when I can just am extremely busy.

olfek:
@mymindstorm ...

> When you use prettier outright instead of through npm run, depending on how your system is configured it might use a different version which may have different rules.

https://prettier.io/docs/en/configuration says

> **Prettier intentionally doesn’t support any kind of global configuration.** This is to make sure that when a project is copied to another computer, Prettier’s behavior stays the same. Otherwise, Prettier wouldn’t be able to guarantee that everybody in a team gets the same consistent results.

olfek:
@mymindstorm ...

I propose temporarily (or permanently?) suspending the Prettier CI step until you set up a config file.

olfek:
> @mymindstorm ...
> 
> After running `prettier` with the `--write` option, I get this from `git diff --stat` :
> 
> ```
>  src/store/Accounts.ts | 79 +++++++++++++++++++++++++--------------------------
>  1 file changed, 38 insertions(+), 41 deletions(-)
> ```
> 
> Would you like me to commit this? - It will pollute the PR diff a little, I await your decision.

@mymindstorm ...

So ... I was getting a big diff because I was using `npx` to run `prettier` like this:

```
npx prettier --check ./src/* ./src/**/* ./src/**/**/* ./sass/*.scss
```

... without running `npm install`, meaning the latest version of prettier (`3.3.3`) was being used instead of the `package.json` declared `2.2.1`.

The default rules have changed since `2.2.1`. See:

https://github.com/prettier/prettier/blob/2.2.1/src/language-js/options.js
https://github.com/prettier/prettier/blob/3.3.3/src/language-js/options.js

https://github.com/prettier/prettier/blob/2.2.1/src/main/core-options.js
https://github.com/prettier/prettier/blob/3.3.3/src/main/core-options.evaluate.js

olfek:
@mymindstorm Can you run CI again?

olfek:
@mymindstorm 📣📣📣

olfek:
@mymindstorm If you're too busy to work on this project, you think maybe it's time to pass it on to someone who can?

mymindstorm:
> @mymindstorm If you're too busy to work on this project, you think maybe it's time to pass it on to someone who can?

Given the sensitive nature of this project, we aren't going to take into account external calls for new maintainers. [We don't want to be the next xz.](https://research.swtch.com/xz-timeline)

Life happens and circumstances require me to put Authenticator on hold outside of high urgency tasks. We appreciate your contributions, and we're working to get them merged as soon as we can.

olfek:
@mymindstorm 

That was an interesting read, I didn't know about it, but with all due respect, there are probably more good cases of handing over the reins than bad, and the potential risks in unmaintained security software has far greater odds I would guess.

olfek:
@mymindstorm @Sneezry 🔔 🔔 🔔

PR #1344, opened 2024-11-24
Update and fix the Hungarian translation

summoner001:
Update and fix the Hungarian translation


summoner001:
Found on crowdin. sorry.

PR #1519, opened 2026-03-13
feat: allow editing HOTP counter in entry edit UI (#1517)

ddjain:
Lets users set or correct the HOTP counter when editing an entry, so it can match the value from another app after migration.

- In edit mode, counter-based (HOTP) entries show an editable "Counter" field.
- Changing it updates the displayed code and persists the value.

Closes #1517 (discussion).


daihaiduong6363-coder:
74

